Issue #87, opened 2020-04-15
ALB Ingress Controller.


errm:
To make the cluster batteries included we should/could ship an ingress controller.

I guess the ALB ingress controller seems like the best AWS native option...

https://github.com/kubernetes-sigs/aws-alb-ingress-controller/

ettiee:
Currently the release tags for `terraform-aws-eks` map to EKS versions. As we are shipping technologies with `terraform-aws-eks` how would we upgrade a version of e.g. ALB Ingress Controller should we want to upgrade those individual technologies?

errm:
> Currently the release tags for terraform-aws-eks map to EKS versions. 

The minor version number can be bumped if/when we want to change something ...

We would only do this though:

* If the upgrade was not a breaking change
* or Included a very important bugfix
* or Included a security fix

The idea is to keep each release fairly stable, and bundle big upgrades of the cluster level components into the k8s version upgrade cycle. Then we only have to worry about handling any upgrade issues on a ~ quarterly basis

errm:
Further to this ... @ettiee have you found a need to upgrade the ALB ingress controller outside of bumping the k8s version?

ettiee:
> Further to this ... @ettiee have you found a need to upgrade the ALB ingress controller outside of bumping the k8s version?

No, but we have found the need to upgrade flux outside of bumping the k8s version

errm:
> No, but we have found the need to upgrade flux outside of bumping the k8s version

I was thinking that flux is not a per cluster component ... so we possibly wouldn't want to manage it in the k8s module anyhow... wdyt?

ettiee:
> > No, but we have found the need to upgrade flux outside of bumping the k8s version
> 
> I was thinking that flux is not a per cluster component ... so we possibly wouldn't want to manage it in the k8s module anyhow... wdyt?

Yeah fine by me, we have it in every cluster atm but other teams might not want it. I'm keen to only put in components which are absolutely required by everyone so that we keep flexibility to choose components within our team if we want to

errm:
The current stratergy is to install everything by default... then provide flags to disable if required:

https://github.com/cookpad/terraform-aws-eks/tree/master/modules/cluster#cluster-critical-add-ons

Does this line up with:

> I'm keen to only have components managed from terraform-aws-eks which are absolutely required by everyone so that we keep flexibility to choose components within our team if we want to

[...]

PR #88, opened 2020-04-21
Adds GPU support

### LINKED ISSUE #59: GPU support

errm:
Fixes #59

### DIFF
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -109,3 +109,9 @@ variable "kms_cmk_arn" {
   default     = ""
   description = "The ARN of the KMS (CMK) customer master key, to be used for Envelope Encryption of Kubernetes secrets, if not set a key will be generated"
 }
+
+variable "nvidia_device_plugin" {
+  type        = bool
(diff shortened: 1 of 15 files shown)


modules/cluster/variables.tf
```diff
@@ -109,3 +109,9 @@ variable "kms_cmk_arn" {
   default     = ""
   description = "The ARN of the KMS (CMK) customer master key, to be used for Envelope Encryption of Kubernetes secrets, if not set a key will be generated"
 }
+
+variable "nvidia_device_plugin" {
+  type        = bool
+  default     = true
```
errm, line 115: I set this to default as `true` since it doesn't seem to cause any harm to try and run on a non-gpu node.

But it kindof ties into https://github.com/cookpad/terraform-aws-eks/issues/87#issuecomment-615109193 perhaps we should default stuff like this to false...

Alternatively perhaps we should just add a label to target the DaemonSet to gpu nodes 🤔

errm: I went down the nodeSlector route!

PR #95, opened 2020-05-05
Use Cluster Security Group - By default

### LINKED ISSUE #94: Make creating the cluster sg optional
By default eks will create a security group for the cluster ... which we can just use.

We should make an option not to create aws_security_group.control_plane and just use this instead!

errm:
Fixes #94

See: https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html for background!

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,0 +1,20 @@
+# Upgrade Notes
+
+## 1.14 -> 1.15
+
+### Cluster Security Group
+
+Existing clusters will be using separately managed security groups for cluster
+and nodes. To continue to use these (and avoid recreating the cluster) set
+`legacy_security_groups = true` on the cluster module.
+
+Update terraform state:
+
+```shell
+export MODULE_NAME=example-cluster
+terraform state mv module.$MODULE_NAME.aws_security_group.control_plane module.$MODULE_NAME.aws_security_group.control_plane[0]
+terraform state mv module.$MODULE_NAME.aws_security_group.node module.$MODULE_NAME.aws_security_group.node[0]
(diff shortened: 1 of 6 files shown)


UPGRADING.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Cluster Security Group
+
+Existing clusters will be using separately managed security groups for cluster
+and nodes. To continue to use these (and avoid recreating the cluster) set
+`legacy_security_groups = true` on the cluster module.
+
+Update terraform state:
+
+```shell
```
mtpereira, line 13: It would be nice to provide this as an actual script, to ease migration and reduce errors.

modules/cluster/variables.tf
```diff
@@ -115,3 +115,9 @@ variable "nvidia_device_plugin" {
   default     = true
   description = "Should the nvidia device plugin be installed, required for GPU nodes"
 }
+
+variable "legacy_security_groups" {
+  type        = bool
+  default     = false
+  description = "Should we create security groups for the cluster and nodes (rather than using the cluster security group), if changed after provisioning - forces the cluster to be recreated"
```
mtpereira, line 122: I'm not sure I understood the wording here, is this what you mean? 
```suggestion
  description = "Should we create security groups for the cluster and nodes (rather than using the cluster security group). If changed after provisioning, it forces the cluster to be recreated"
```
Is it easier to understand like this?

UPGRADING.md
errm: Good idea!

modules/cluster/variables.tf
errm: agreed this is pretty lousy ... hopefully this is a little better!

errm:
> :shipit:

When the tests pass :trollface:

mtpereira:
> > :shipit:
> 
> When the tests pass :trollface:

They are still running :hand_over_mouth:

PR #99, opened 2020-05-05
Adds rules to allow traffic between legacy "node" sg and cluster sg

errm:
This allows communication to continue between nodes with the legacy sg,
and those using the cluster sg. And means we can remove the legacy node
sg in a later update!

### DIFF
--- a/modules/cluster/security_groups.tf
+++ b/modules/cluster/security_groups.tf
@@ -24,6 +24,8 @@ resource "aws_security_group" "control_plane" {
 /*
   Nodes Security Group
   And rules to control communication between nodes and cluster.
+
+  
 */
 
 resource "aws_security_group" "node" {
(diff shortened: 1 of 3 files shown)


modules/cluster/security_groups.tf
```diff
@@ -69,26 +71,38 @@ resource "aws_security_group_rule" "node_ingress_self" {
...
+  description              = "Allow worker Kubelets and pods to receive communication from the cluster control plane"
+  from_port                = 0
+  to_port                  = 65535
   protocol                 = "tcp"
-  security_group_id        = aws_security_group.control_plane[count.index].id
-  source_security_group_id = aws_security_group.node[count.index].id
+  security_group_id        = aws_security_group.node[count.index].id
+  source_security_group_id = aws_eks_cluster.control_plane.vpc_config.0.cluster_security_group_id
   type                     = "ingress"
 }
```
mtpereira, line 108: Aren't these two resources redundant? They both enable comm from the control plane to the nodes and the first allows all protocols instead of just TCP. It feels like the second one should be removed. Am I missing something? Thanks. :bow:

errm: yes you are correct.

PR #100, opened 2020-05-06
Remove legacy security groups

errm:
The node security group should no longer be used by node groups created
by the 1.15 version of the module. So it should be safe to remove in
1.16


mtpereira:
So this change is for the 1.16 release of this module?

errm:
> So this change is for the 1.16 release of this module?

Yes correct

PR #107, opened 2020-05-13
Update node exporter in line with documentation

### LINKED ISSUE #106: Fix node-exporter config
To follow: https://github.com/prometheus/node_exporter/tree/v0.18.1#using-docker

errm:
Fixes #106

### DIFF
--- a/modules/cluster/addons/prometheus-node-exporter.yaml
+++ b/modules/cluster/addons/prometheus-node-exporter.yaml
@@ -8,7 +8,6 @@ metadata:
   labels:     
     app: prometheus-node-exporter
     release: prometheus-node-exporter
-    chart: prometheus-node-exporter-1.9.1
 spec:
   privileged: false
   # Required to prevent escalations to root.
@@ -60,7 +59,6 @@ metadata:
   namespace: kube-system
   labels:
     app: prometheus-node-exporter
-    chart: prometheus-node-exporter-1.9.1
     release: "prometheus-node-exporter"
 imagePullSecrets:
   []
@@ -73,7 +71,6 @@ metadata:
   labels:     
     app: prometheus-node-exporter
     release: prometheus-node-exporter
-    chart: prometheus-node-exporter-1.9.1
 rules:
 - apiGroups: ['extensions']
   resources: ['podsecuritypolicies']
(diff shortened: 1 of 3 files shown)


modules/cluster/addons/prometheus-node-exporter.yaml
```diff
@@ -157,8 +150,8 @@ spec:
           image: "prom/node-exporter:v0.18.1"
           imagePullPolicy: IfNotPresent
           args:
-            - --path.procfs=/host/proc
-            - --path.sysfs=/host/sys
+            - --collector.filesystem.ignored-mount-points=^/(dev|var/lib/docker|proc|sys|var/lib/kubelet)($|/)
```
mtpereira, line 153: Really just a nit but unless for some reason the current regex is more efficient than the following, I'd suggest we'd use this instead, since it is a more common "idiom" and a bit easier to parse.
```suggestion
            - --collector.filesystem.ignored-mount-points=^/(dev|var/lib/docker|proc|sys|var/lib/kubelet)?/$
```

mtpereira: Oh I guess I was too slow!... :sweat:

errm: noted ...

kaorimatz: I don't think that's what we want here. It doesn't match for example a mount point like /var/lib/kubelet/pods/path/to/volume.

mtpereira: It matches the exact same thing it matched before AFAIK. I see what you mean though, but if you're ignoring `/var/lib/kubelet`, which is covered by the regex, wouldn't your example be ignored as well implicitly?

kaorimatz: The current regular expression lets us ignore a mount point like `/var/lib/kubelet/pods/path/to/volum`, but you regular expression doesn't. It only matches `/var/lib/kubelet` or `/var/lib/kubelet/`.

mtpereira: Oh I see the misunderstanding here, this `collector.filesystem.ignored-mount-points` option does not enforce the resulting string to be the full path of the mount, so if there is no string termination it will match anything that could be appended to the path. Got it.

PR #108, opened 2020-05-19
Prepare for prometheus

errm:
* Change some defaults to not deploy metrics components by default
* Pass dns-cluster-ip to coredns and nodes


errm:
> I just wonder if we could add any tests for these changes?

Yes... yes we should!

PR #110, opened 2020-05-20
Update Amazon VPC CNI plugin to 1.6.1

kaorimatz:
The manifest is updated with the following command:

```
curl -s https://raw.githubusercontent.com/aws/amazon-vpc-cni-k8s/release-1.6.1/config/v1.6/aws-k8s-cni.yaml -o modules/cluster/addons/aws-k8s-cni.yaml
```

1.6.1 is the version currently recommended in https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html.

Changelog: https://github.com/aws/amazon-vpc-cni-k8s/blob/master/CHANGELOG.md#v161

### DIFF
--- a/modules/cluster/addons/aws-k8s-cni.yaml
+++ b/modules/cluster/addons/aws-k8s-cni.yaml
@@ -77,7 +77,20 @@ spec:
                     operator: In
                     values:
                       - amd64
-                  - key: eks.amazonaws.com/compute-type
+                  - key: "eks.amazonaws.com/compute-type"
+                    operator: NotIn
+                    values:
+                      - fargate
+              - matchExpressions:
+                  - key: "kubernetes.io/os"
+                    operator: In
+                    values:
+                      - linux
+                  - key: "kubernetes.io/arch"
+                    operator: In
+                    values:
+                      - amd64
+                  - key: "eks.amazonaws.com/compute-type"
                     operator: NotIn
                     values:
                       - fargate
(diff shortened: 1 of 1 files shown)


modules/cluster/addons/aws-k8s-cni.yaml
```diff
@@ -86,15 +99,27 @@ spec:
...
+            exec:
+              command: ["/app/grpc-health-probe", "-addr=:50051"]
+            initialDelaySeconds: 35
           env:
             - name: AWS_VPC_K8S_CNI_LOGLEVEL
               value: DEBUG
+            - name: AWS_VPC_K8S_CNI_VETHPREFIX
+              value: eni
+            - name: AWS_VPC_ENI_MTU
+              value: "9001"
```
mtpereira, line 122: Why are we defining this value, given that we're using the default value anyway? https://github.com/aws/amazon-vpc-cni-k8s/blob/501ca48d682670c45e8303c631cec8e178db6c89/pkg/networkutils/network.go#L935

Maybe it would be better to have us use the default value, which presumably would be the recommended value for what is the most common setup (linux, amd64) for this tool?

errm: I think setting the default here is OK ... it makes it more explicit so we can see what it is without checking the sourcecode.

Also it's best not to change the yaml from the upstream if we can help it ... it makes updates simpler!

mtpereira: Oh I didn't realize this was the upstream manifest. Cool let's keep it. :smiley:

kaorimatz:
Thank you!

PR #113, opened 2020-05-21
fix state bug in 1.15 upgrade

ettiee:


errm:
a) Let’s upgrade another cluster to check this fixes the process before we merge...
b) The tests are failing because of how github actions doesn’t let the secrets be accessed from a fork... this script isn’t tested so no matter.

ettiee:
> a) Let’s upgrade another cluster to check this fixes the process before we merge...

Yup was going to suggest that this morning - I have one more to upgrade today so I'm going to test this

PR #116, opened 2020-06-05
Add IAM permissions_boundary variable for cluster_autoscaler IAM role

takanabe:
## Background

I'd like to add an option to set up [IAM Permissions boundary](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html)

A permissions boundary is an advanced feature for using a managed policy to set the maximum permissions that an identity-based policy can grant to an IAM entity. An entity's permissions boundary allows it to perform only the actions that are allowed by both its identity-based policies and its permissions boundaries.

## Reference

* Terrafrom aws_iam_role: https://www.terraform.io/docs/providers/aws/d/iam_role.html


takanabe:
Let me push the same PR from a none-forked repo 📝

PR #117, opened 2020-06-05
Add IAM permissions_boundary variable for cluster_autoscaler IAM role

takanabe:
## Background

I'd like to add an option to set up [IAM Permissions boundary](https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies_boundaries.html)

A permissions boundary is an advanced feature for using a managed policy to set the maximum permissions that an identity-based policy can grant to an IAM entity. An entity's permissions boundary allows it to perform only the actions that are allowed by both its identity-based policies and its permissions boundaries.

## Reference

* Terrafrom aws_iam_role: https://www.terraform.io/docs/providers/aws/d/iam_role.html

### DIFF
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -42,7 +42,7 @@ variable "iam_config" {
   description = "The IAM roles used by the cluster, If you use the included IAM module you can provide it's config output variable."
 }
 
-variable "iam_permissions_boundary" {
+variable "cluster_autoscaler_iam_permissions_boundary" {
   type        = string
   default     = ""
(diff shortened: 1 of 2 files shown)


modules/cluster/variables.tf
```diff
@@ -42,6 +42,12 @@ variable "iam_config" {
   description = "The IAM roles used by the cluster, If you use the included IAM module you can provide it's config output variable."
 }
 
+variable "iam_permissions_boundary" {
```
errm, line 45: I think this needs a more specific name:
```suggestion
variable "cluster_autoscaler_iam_permissions_boundary" {
```

PR #119, opened 2020-06-19
Add variables for critical-addons

takanabe:
## Background

It's handy if we can change the configurations of the `critical-addon` node group and `aws_autoscaling_group` based on purposes to use EKS clusters. For example, I don't need huge instances for the `critical-addons` node group when EKS clusters are used for research purposes and want to change the value of `mix_size` and `max_size` for the `asg_node_group` module.

## Changes

* Add variables for `critical-addon` asg_node_group to the cluster module

### DIFF
--- a/modules/cluster/addons.tf
+++ b/modules/cluster/addons.tf
@@ -7,7 +7,6 @@ module "critical_addons_node_group" {
   key_name         = var.critical_addons_node_group_key_name
   min_size         = var.critical_addons_node_group_min_size
   max_size         = var.critical_addons_node_group_max_size
-  desired_capacity = var.critical_addons_node_group_desired_capacity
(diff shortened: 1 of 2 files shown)


modules/asg_node_group/main.tf
```diff
@@ -106,6 +106,7 @@ resource "aws_autoscaling_group" "nodes" {
   name                      = "${local.name_prefix}-${each.key}"
   min_size                  = local.min_size
   max_size                  = local.max_size
+  desired_capacity          = var.desired_capacity
```
errm, line 109: I don't think that we want to set desired_capacity.

cluster_autoscaler is responsible for setting this.

If cluster_autoscaler increases this value then we run terraform we could cause unexpected scale in!

PR #121, opened 2020-06-22
Open metrics port on kube-proxy

errm:


modules/cluster/addons/kube-proxy.yaml
```diff
@@ -32,7 +32,7 @@ spec:
       - command:
         - /bin/sh
         - -c
-        - kube-proxy --v=2 --config=/var/lib/kube-proxy-config/config
+        - kube-proxy --v=2 --config=/var/lib/kube-proxy-config/config --metrics-bind-address=0.0.0.0
```
dan-slinky-ckpd, line 35: I think `--metrics-bind-address` needs the port too, so `--metrics-bind-address=0.0.0.0:10249`

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-proxy/config/v1alpha1/types.go#L117

PR #130, opened 2020-08-25
Adds the ignoreDaemonSets value to the helm config for aws-node-termination-handler

errm:
This option was already changed in the generated manifest in #126 this
change just avoids undoing that change when we run `hack/generate_addons.sh`

FYI: @takanabe


takanabe (approved):
oh, the manifest was generated via Helm! Thank you for the follow-up!

PR #136, opened 2020-09-03
ALB Ingress Controller

### LINKED ISSUE #87: ALB Ingress Controller.

errm:
This PR introduces support for provisioning the ALB ingress controller.

For now the default is to not provision this component, based on the discussions in #87 as some cluster admins might prefer to use another implementation.

It can be enabled by setting `aws_alb_ingress_controller = true` on the Cluster module.

Fixes #87

### DIFF
--- a/.github/actions/terratest/entrypoint.sh
+++ b/.github/actions/terratest/entrypoint.sh
@@ -1,4 +1,6 @@
-#!/bin/sh -l
+#!/bin/bash
+
+set -euo pipefail
 
 # If the terraform version is specified install the correct one
 tfenv install || true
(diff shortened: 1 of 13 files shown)


errm:
I copied this from the terratest documentation https://terratest.gruntwork.io/docs/getting-started/quick-start/

In all the examples they do this... https://github.com/gruntwork-io/terratest/search?q=http-helper&unscoped_q=http-helper they also follow the same pattern when importing `test-structure` https://github.com/gruntwork-io/terratest/search?q=test-structure&unscoped_q=test-structure

I am not sure if it is idiomatic, as I think the real go idiom is to not have dashes in package names https://blog.golang.org/package-names, I this case I am happy to just stay consistent to what they are doing in the tereatest examples, since we don't get to change there package naming!

mtpereira (approved):

PR #140, opened 2020-09-11
Adds node region and zone topology labels to autoscaler template

errm:
This will assist cluster autoscaler in choosing the right ASG to scale
up when zone is important, e.g. when handling EBS volumes or zone
(anti)affinity!

### DIFF
--- a/modules/asg_node_group/main.tf
+++ b/modules/asg_node_group/main.tf
@@ -35,6 +35,8 @@ data "aws_ami" "image" {
   }
 }
 
+data "aws_region" "current" {}
+
 data "template_file" "cloud_config" {
   template = file("${path.module}/cloud_config.tpl")
   vars = {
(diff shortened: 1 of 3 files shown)


mtpereira:
I couldn't find the documentation for this anywhere. :disappointed: Do you have a URL for that? Thank you!

errm:
Documentation about hinting the autoscaler which labels the nodes will have when they join the cluster is here: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#auto-discovery-setup

Topology labels are documented here https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone

https://github.com/kubernetes-sigs/aws-ebs-csi-driver mentions topology.ebs.csi.aws.com/zone

PR #146, opened 2020-09-26
Fix flaky cleanup

### LINKED ISSUE #144: Fix issue where CI doesn't correctly teardown the testing environment.
We are seeing common CI fails where terraform fails to cleanup the VPC because a security group is still present!

errm:
Fixes #144

### DIFF
--- a/examples/cluster/environment/outputs.tf
+++ b/examples/cluster/environment/outputs.tf
@@ -1,0 +1,3 @@
+output "vpc_id" {
+  value = module.vpc.config.vpc_id
+}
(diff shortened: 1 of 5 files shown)


mtpereira:
I think I recall dealing with this in the past and my conclusion was that it can't really be done with 100% success, since sometimes the SG removal doesn't get completed even with an explicit request for removal. :(

errm:
The issue is that when we request for the EKS cluster to be deleted, it *should* remove the security group(s) that it adds to the VPC.

For some reason this doesn't happen 100% of the time.

mtpereira:
@errm Right, my comment was regarding the fix approach: even with explicit delete requests to those SG that should no longer be there, they sometimes still linger.

PR #150, opened 2020-09-30
Add kube-bench

errm:
Use kube-bench https://github.com/aquasecurity/kube-bench to run
security benchmarks against nodes.

### DIFF
--- a/test/cluster_test.go
+++ b/test/cluster_test.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -89,6 +90,7 @@ func TestTerraformAwsEksCluster(t *testing.T) {
 		defer cleanupTerraform(t, nodeGroupDir)
 		validateNodeTerminationHandler(t, kubeconfig)
 		validateClusterAutoscaler(t, kubeconfig)
+		validateKubeBench(t, kubeconfig)
 		validateStorage(t, kubeconfig)
 		validateIngress(t, kubeconfig)
 		overideAndApplyTerraform(t, workingDir, map[string]interface{}{
@@ -105,6 +107,7 @@ func TestTerraformAwsEksCluster(t *testing.T) {
 		deployTerraform(t, gpuNodeGroupDir, map[string]interface{}{})
 		defer cleanupTerraform(t, gpuNodeGroupDir)
 		validateGPUNodes(t, kubeconfig)
+		validateKubeBench(t, kubeconfig)
 		validateNodeTerminationHandler(t, kubeconfig)
 	})
 }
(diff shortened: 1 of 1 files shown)


test/cluster_test.go
```diff
@@ -445,3 +448,74 @@ spec:
...
+  template:
+    metadata:
+      labels:
+        app: kube-bench
+    spec:
+      hostPID: true
+      containers:
+        - name: kube-bench
+          image: aquasec/kube-bench:latest
+          command: ["kube-bench", "node", "--benchmark", "eks-1.0", "--json"]
```
mtpereira, line 496: Perhaps we should drop the `"node"` argument and let all kube-bench run on all `eks-1.0` targets? https://github.com/aquasecurity/kube-bench#running-kube-bench 
Is there any reason to run this against nodes only?
```suggestion
          command: ["kube-bench", "--benchmark", "eks-1.0", "--json"]
```

errm: kube-bench has to run on the node that it is testing, on EKS we can't schedule workloads to run on the masters since they are managed.

Also as users of EKS we have to trust AWS to keep the control plane secure.

The manifest here is very closely based on https://github.com/aquasecurity/kube-bench/blob/master/job-eks.yaml and the instructions here https://github.com/aquasecurity/kube-bench#running-in-an-eks-cluster

The only difference is that I don't feel it is necessary to build our own container and push it to a ECR repo, since it will only ever be run on an isolated test cluster.

mtpereira: The `eks-1.0` CIS Benchmark does not run on the masters, please check the table here: https://github.com/aquasecurity/kube-bench#running-kube-bench . It targets controlplane, node, policies and managedservices; it does not target master nodes nor etcd.

errm: Right... we are not specifying the `--targets` flag here, so the command we run will run the benchmarks for that list.

The sub command node, just lets kubebench know that it is running on a node, rather than a master.

If we run this command:

`["kube-bench", "--benchmark", "eks-1.0", "--json"]`

Everything works, but we get the following warnings in the log output:

```
W1007 12:31:31.906369   23804 util.go:96]
Unable to detect running programs for component "apiserver"
The following "master node" programs have been searched, but none of them have been found:
	- kube-apiserver
	- hyperkube apiserver
	- hyperkube kube-apiserver
	- apiserver


These program names are provided in the config.yaml, section 'master.apiserver.bins'
```
It looks like running kube-bench in this way generates a lot of "Not Scored" results, this means that the checks couldn't run, or are manual!

I am fairly sure this is the correct way to check the node configuration https://github.com/aquasecurity/kube-bench#running-in-an-eks-cluster

mtpereira: I understand now, thank you! :bow:

PR #173, opened 2021-02-03
fix ci checkout action

dan-slinky-ckpd:
They've renamed their `master` to `main`. Could use `v2` tag as well 🤷

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout'
-        uses: actions/checkout@master
+        uses: actions/checkout@main
       - name: 'Terraform Format'
         uses: hashicorp/terraform-github-actions@master
         with:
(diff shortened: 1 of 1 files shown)


danslinky:
Note to future @dan-slinky-ckpd - the json is not being escaped correctly, e.g. ` \"nodefs.available\": \"10%!\(MISSING)",\n    \"nodefs.inodesFre`

.github/workflows/ci.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: 'Checkout'
-        uses: actions/checkout@master
+        uses: actions/checkout@main
       - name: 'Terraform Format'
         uses: hashicorp/terraform-github-actions@master
```
ettiee, line 19: I can see this has been superseded by https://github.com/hashicorp/setup-terraform as well ✍️ perhaps some more cleanup in another task

dan-slinky-ckpd:
> What repository has renamed to `main`? I think this action will check out this repository which has head on master ?

https://github.com/actions/checkout

ettiee:
> > What repository has renamed to `main`? I think this action will check out this repository which has head on master ?

Ah I see - maybe we can checkout a version rather than a branch to prevent breaking changes

ettiee (approved):

aidy (approved):

Issue #175, opened 2021-02-08
Fix flaky kubebench test


ettiee:
See here for an example of problematic json https://github.com/cookpad/terraform-aws-eks/actions/runs/535074180

aidy:
Fixed with #178

PR #178, opened 2021-02-08
Update tests for new kube-bench format

aidy:
The format of kube-bench output changed from 0.4.0 to 0.5.0.

DriveBy: Fix json tag names - this might result in *different* failures
as everything might not get the default 0 value.


aidy:
connects #175

This does fix the parsing of kube-bench results, but not-unexpectedly, we're seeing failures as a result of that.

This moves in the right direction, and is less broken than before.

ettiee (approved):
LGTM - we can merge and work on the fix for the kube-bench tests then 👍

PR #187, opened 2021-02-12
[ASG Node Group] Make EKS node name self-explanatory

takanabe:
## Background 

EC2 instances used as EKS critical addons and workers have the following
name conventions.

```
eks-node-{cluster_name}-{instance_id}
```

and actual EC2 names are like

```
aws --region us-west-2 ec2 describe-instances --output json --filter "Name=instance-state-name,Values=running" --query 'Reservations[].Instances[].[Tags[?Key==`Name`].Value]'
[
    [
        [
            "eks-node-staging-i-0f0f023b78914d96d"
        ]
    ],
    [
        [
            "eks-node-staging-i-0bfd628ea84f245c1"
        ]
    ],
    [
        [
            "eks-node-staging-i-04e842972094f606c"
        ]
    ],
    [
        [
            "eks-node-staging-i-02b89112b0d523510"
        ]
    ]
]

```

In theory, we should not name serves(pets vs cattle) to identify the servers but adding metadata to EC2 instance names are sometimes useful. For example, the following names explain the type of EKS node and not the role of applications running in the node.

```
eks-{cluster_name}-{asg_node_group_name}-{instance_family}-{instance_size}-{instance_lifecycle}-{instance_id}
```
The expected EC2 name will become as follows:

```
aws --region us-west-2 ec2 describe-instances --output json --filter "Name=instance-state-name,Values=running" --query 'Reservations[].Instances[].[Tags[?Key==`Name`].Value]'

[
    [
        [
            "eks-staging-worker-burstable-nano-spot-i-09a7dc7273050d3d2"
        ]
    ],
    [
        [
            "eks-staging-worker-burstable-nano-spot-i-068db7c80fd94acc4"
        ]
    ],
    [
        [
            "eks-staging-critical-addons-burstable-small-spot-i-0a2ceda5188392458"
        ]
    ],
    [
        [
            "eks-staging-worker-burstable-nano-spot-i-0e3f3e9e16eb0b7f3"
        ]
    ]
]

```

I thought the above metadata is good to add but if we don't think they are not useful. I'd like to add metadata `worker` and `critical-addons` at least.

## Change

* Remove `node` from EC2 Name tag to make the name shorter
   * since the strings does not include valuable info
* EC2 name tags will change 
   * from `eks-node-${cluster_name}-{{ v1.instance_id }}`
   * to `eks-{cluster_name}-{asg_node_group_name}-{instance_family}-{instance_size}-{instance_lifecycle}-{instance_id}`

## Question

This change may impact the design of this module and want to get feedback from maintainers.

### DIFF
--- a/modules/asg_node_group/cloud_config.tpl
+++ b/modules/asg_node_group/cloud_config.tpl
@@ -1,7 +1,7 @@
 ## template: jinja
 #cloud-config
-fqdn: eks-node-${cluster_name}-{{ v1.instance_id }}
+fqdn: ${node_name_prefix}-{{ v1.instance_id }}
 runcmd:
-- [aws, --region={{ v1.region }}, ec2, create-tags, --resources={{ v1.instance_id }}, "--tags=Key=Name,Value=eks-node-${cluster_name}-{{ v1.instance_id }}"]
+- [aws, --region={{ v1.region }}, ec2, create-tags, --resources={{ v1.instance_id }}, "--tags=Key=Name,Value=${node_name_prefix}-{{ v1.instance_id }}"]
 - [systemctl, restart, docker]
 - [/etc/eks/bootstrap.sh, ${cluster_name}, --kubelet-extra-args, '--node-labels=${labels} --register-with-taints="${taints}"', --dns-cluster-ip, ${dns_cluster_ip}]
--- a/modules/asg_node_group/main.tf
+++ b/modules/asg_node_group/main.tf
@@ -8,7 +8,7 @@ locals {
 
   instance_types       = length(var.instance_types) > 0 ? var.instance_types : [for instance_family in local.preset_instance_families[var.instance_family] : "${instance_family}.${var.instance_size}"]
   instance_overrides   = var.instance_lifecycle == "spot" ? local.instance_types : [local.instance_types[0]]
-  name_prefix          = replace(join("-", compact(["eks-node", var.cluster_config.name, var.name, var.instance_family, var.instance_size, var.instance_lifecycle])), "_", "-")
+  name_prefix          = replace(join("-", compact(["eks", var.cluster_config.name, var.name, var.instance_family, var.instance_size, var.instance_lifecycle])), "_", "-")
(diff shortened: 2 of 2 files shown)


errm:
I don't have a strong opinion about what the node names should be, but have a few thoughts:

* Perhaps we could just make each of these metadatas seperate tags on the instances.
* Perhaps we could try and make the `Name` tag align with the names in kubernetes from `kubectl get nodes` etc... this makes it simpler to jump from looking at a container in kubernetes to ssh to the node for some debug reasons.
* Perhaps these metadatas are not really that useful? Most operations e.g. rolling out a new AMI can just be handled at the ASG level.
* Currently nodes using bottlerocket don't get unique names since we can't just run a script on boot, they just all have the name from the ASG. If we think unique names are important then we should fix this (with a deamonset or somesuch), if we don't think that unique names are important, then lets just remove this feature for all host oses for consistency.

mtpereira:
:+1: on tags over names!

ettiee:
Some historical context - I think the metadata node naming conventions was a feature request from other teams so it might be useful to ask module users if / why this feature is useful to them before we make a decision on next steps

mattjw:
I'm don't have a passionate view on node naming.

> In theory, we should not name serves(pets vs cattle) to identify the servers

Agree.

>  but adding metadata to EC2 instance names are sometimes useful

Also agree – it can help give additional context about an EKS nodes at-a-glance.

Maybe "cattle not pets" and "human-readable, glanceable node names" are not mutually exclusive?

> Perhaps we could try and make the Name tag align with the names in kubernetes from kubectl get nodes etc... this makes it simpler to jump from looking at a container in kubernetes to ssh to the node for some debug reasons.

This is a nice idea also.

ettiee:
@takanabe we discussed this in the k8s user group - and the consensus agreed with:

> Perhaps we could just make each of these metadatas seperate tags on the instances.

but also nobody held a strong opinion on this

takanabe:
Let me close this issue since the metadata instead of node name changes makes sense to me but it's beyond this PR!

PR #185, opened 2021-02-11
[IAM module] Fix Invalid depends_on reference error

takanabe:
## Background

The following error occurs while terraform plan is executed.

<details>
<summary>terraform plan</summary>

```
terraform plan
Acquiring state lock. This may take a few moments...
Releasing state lock. This may take a few moments...

Error: Invalid depends_on reference

  on .terraform/modules/eks_iam/modules/iam/outputs.tf line 7, in output "config":
   7:     aws_iam_role_policy_attachment.eks_worker_node.policy_arn,

References in depends_on must be to a whole object (resource, etc), not to an
attribute of an object.

Error: Invalid depends_on reference

  on .terraform/modules/eks_iam/modules/iam/outputs.tf line 8, in output "config":
   8:     aws_iam_role_policy_attachment.eks_cni.policy_arn,

References in depends_on must be to a whole object (resource, etc), not to an
attribute of an object.

Error: Invalid depends_on reference

  on .terraform/modules/eks_iam/modules/iam/outputs.tf line 9, in output "config":
   9:     aws_iam_role_policy_attachment.ecr.policy_arn,

References in depends_on must be to a whole object (resource, etc), not to an
attribute of an object.

Error: Invalid depends_on reference

  on .terraform/modules/eks_iam/modules/iam/outputs.tf line 10, in output "config":
  10:     aws_iam_role_policy_attachment.eks_service_policy.policy_arn,

References in depends_on must be to a whole object (resource, etc), not to an
attribute of an object.

Error: Invalid depends_on reference

  on .terraform/modules/eks_iam/modules/iam/outputs.tf line 11, in output "config":
  11:     aws_iam_role_policy_attachment.eks_cluster_policy.policy_arn,

References in depends_on must be to a whole object (resource, etc), not to an
```
</details>

To fix this invalid dependency errors, we have to pass objects instead of resouces(policy_arn.)

## Technical changes

* Pass objects instead of resouces(policy_arn) to `depends_on` argument.

## Test Environment

A clean AWS account with the following Terraform configurations.

```
$ cat main.tf
module "eks_iam" {
  source  = "cookpad/eks/aws//modules/iam"
  version = "1.17.0"
}

$ cat .terraform-version
0.14.5
```

## Terraform plan with this PR

<details>
<summary>terraform plan</summary>

```
$ terraform plan
Acquiring state lock. This may take a few moments...

An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

[...]

PR #188, opened 2021-02-12
CI failure investigation

ettiee:
Poked about a bit testing locally, some ASGs appeared to not have started in a few runs. I then applied the example terraform configuration piece by piece to see what was going on, and hit
```
Error: InvalidLaunchTemplateName.AlreadyExistsException: Launch template name already in use.
```
I _think_ that the by not having unique names for the node groups we're causing this error... this would perhaps explain the CI failures we're seeing - one of the node groups isn't able to start, which would result in some failures in test workloads because some tainted node is missing / unable to start up...
Can't see the `InvalidLaunchTemplateName` in the CI output but it might be being squashed somewhere, need to investigate that piece further..


ettiee:
Re-ran tests on https://github.com/cookpad/terraform-aws-eks/pull/185 without these changes and got ✅ 
Currently adding some helpers so we can upload some resource stats (pods/nodes etc) as artifacts with https://github.com/actions/upload-artifact - this will give a bit more context to the CI results

errm:
This sounds like it could be some sort of race condition.

I think in the tests we normally destroy the standard node group asg before we start provisioning the other one, but perhaps if things are a little async a launch template can hang around for a bit too long after terraform thinks it is destroyed.

I don't think making this change can break anything so lets merge it!

PR #191, opened 2021-02-17
[addons]  Use public NVIDIA image repository for nvidia-device-plugin

takanabe:
## Background

To bypass the Dockerhub rate limits for the nvidia-device-plugin image, I'd like to use the public image registry which NVIDIA manages.

This PR also includes a [small change](https://github.com/cookpad/terraform-aws-eks/commit/38af30601b44f535da1be8f5a5dcc46902c2db9f) for `aws-ebs-csi-driver` which was generated by executing `hack/generate_addons.sh`.

### DIFF
--- a/hack/generate_addons.sh
+++ b/hack/generate_addons.sh
@@ -19,7 +19,7 @@ kustomize_build() {
 
 helm_template eks aws-node-termination-handler 0.13.3
 helm_template autoscaler cluster-autoscaler 1.0.4 -chart
-helm_template nvdp nvidia-device-plugin 0.7.0
+helm_template nvdp nvidia-device-plugin 0.8.2
 
 curl -o $ADDONS_DIR/kustomize/overlays/metrics-server/resources/metrics-server.yaml -L https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.3.7/components.yaml
 kustomize_build metrics-server
--- a/modules/cluster/addons/nvidia-device-plugin.yaml
+++ b/modules/cluster/addons/nvidia-device-plugin.yaml
@@ -19,10 +19,10 @@ metadata:
   name: nvidia-device-plugin
   namespace: kube-system
   labels:
-    helm.sh/chart: nvidia-device-plugin-0.7.0
+    helm.sh/chart: nvidia-device-plugin-0.8.2
     app.kubernetes.io/name: nvidia-device-plugin
     app.kubernetes.io/instance: nvidia-device-plugin
-    app.kubernetes.io/version: "0.7.0"
+    app.kubernetes.io/version: "0.8.2"
 spec:
   selector:
     matchLabels:
@@ -48,7 +48,7 @@ spec:
       securityContext:
         {}
       containers:
-      - image: nvidia/k8s-device-plugin:v0.7.0
+      - image: nvcr.io/nvidia/k8s-device-plugin:v0.8.2
         imagePullPolicy: IfNotPresent
         name: nvidia-device-plugin-ctr
         args:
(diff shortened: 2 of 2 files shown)


takanabe:
The change https://github.com/cookpad/terraform-aws-eks/commit/38af30601b44f535da1be8f5a5dcc46902c2db9f does not seem better than before. I personally think GCR is much more reliable than quay. Let me check it out again.

modules/cluster/addons/aws-ebs-csi-driver.yaml
```diff
@@ -924,7 +924,7 @@ spec:
         env:
         - name: CSI_ENDPOINT
           value: unix:/csi/csi.sock
-        image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:v0.9.0
+        image: k8s.gcr.io/provider-aws/aws-ebs-csi-driver:latest
```
dan-slinky-ckpd, line 927: The use of the `latest` tag caught my eye here. The `latest` image tag is built from the `master` branch of https://github.com/kubernetes-sigs/aws-ebs-csi-driver/

I note we're referencing the `alpha` overlay [here](https://github.com/cookpad/terraform-aws-eks/blob/master/modules/cluster/addons/kustomize/overlays/aws-ebs-csi-driver/kustomization.yaml#L4)

This includes `base` which does specify `latest`.

If we update our kustomization to use the `stable` overlay, we'll return to using the `v0.9.0` tag, as defined [here](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/v0.9.0/deploy/kubernetes/overlays/stable/kustomization.yaml#L7)

takanabe: >The use of the latest tag caught my eye here. 

Yes, I have the same opinion and I don't want to use the latest tag in this module.

Using `stable` overlay is the best choice if we don't support the `alpha` feature 

>Starting from Kubernetes 1.17, CSI migration is supported as beta feature (alpha since 1.14). If you have persistence volumes that are created with in-tree kubernetes.io/aws-ebs plugin, you could migrate to use EBS CSI driver. To turn on the migration, set CSIMigration and CSIMigrationAWS feature gates to true for kube-controller-manager and kubelet.

takanabe: @dan-slinky-ckpd 

[Accroding to the official document](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html), EKS 1.17 does not support the CSI migration feature and the later version didn't say they support the feature.

>EKS has not enabled the CSIMigrationAWS feature flag. This will be enabled in a future release, along with detailed migration instructions. For more info on CSI migration, see the Kubernetes blog.

So, +1 to the stable + `v0.9.0` tag. WDYT?

dan-slinky-ckpd: 👍 `stable`

errm: when I added this to the module we were on alpha because beta didn't support volume resizing.

It would be good to check the status on that!

errm: Ok... volume resizing is beta since 1.16 so LGTM!

takanabe:
Let me focus on the image update of `nvidia-device-plugin` in this PR. I will handle the update of `aws-ebs-csi-driver` with https://github.com/cookpad/terraform-aws-eks/pull/196

dan-slinky-ckpd (approved):

takanabe:
Thanks!

Issue #171, opened 2021-01-26
Allow users to pass their own container registry as a variable


ettiee:
Allowing them to avoid docker rate limits if desired

ettiee:
Is a current bug, but if the upstream projects have the images hosted elsewhere (not dockerhub) we could just update those instead

ettiee:
For the 1.18 release we will remove any remaining dependency on dockerhub (e.g. #191 ) but won't include this feature.
But we'll keep this issue open so we can add it later

errm:
All add-ons installed by the module use public eks images managed by aws now!

PR #194, opened 2021-02-24
pull prometheus node-exporter from quay.io

ettiee:
To remove dependency on dockerhub


ettiee:
See https://github.com/cookpad/terraform-aws-eks/issues/171#issuecomment-784957558

aidy (approved):
lgtm

PR #199, opened 2021-02-26
[Revisit v0.20.0][nits] Fix hacks roll_nodes.sh

kenju:
- https://github.com/cookpad/terraform-aws-eks/commit/5a1ff54a43b9581eb286bddc526109fdea0e939d
- https://github.com/cookpad/terraform-aws-eks/commit/c474fcbade4a5d7d5d041ac3b6bc04acbda95794

### DIFF
--- a/hack/roll_nodes.sh
+++ b/hack/roll_nodes.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# roll_nodes.sh TARGET_VERSION
+# roll_nodes.sh
 #
 # Replace Kubernetes Nodes, to roll out config changes
 #
--- a/hack/roll_nodes.sh
+++ b/hack/roll_nodes.sh
@@ -72,7 +72,7 @@ done
 for NODE in $NODES_TO_ROLL
 do
   detach $NODE
-  kubectl drain --delete-local-data --ignore-daemonsets $NODE
+  kubectl drain --delete-emptydir-data --ignore-daemonsets $NODE
   aws $AWS_EXTRA_ARGS ec2 terminate-instances --instance-ids $(instance_id $NODE)
   wait_for_pods
 done


hack/roll_nodes.sh
```diff
@@ -72,7 +72,7 @@ done
 for NODE in $NODES_TO_ROLL
 do
   detach $NODE
-  kubectl drain --delete-local-data --ignore-daemonsets $NODE
+  kubectl drain --delete-emptydir-data --ignore-daemonsets $NODE
```
demijonathon, line 75: I can't find this as a flag. The only option I've seen is --delete-local-data=false ?
https://github.com/kubernetes/kubernetes/issues/80228
Can you explain what this is doing (in the PR).

kenju: There is one PR linked in that issue

https://github.com/kubernetes/kubernetes/pull/95076

Basically, only the name of this flag is changed. What it does is unchanged. The original flag (`--delete-local-data`) seems to be misleading, that's why the new flag `--delete-emptydir-data` was introduced.

demijonathon (approved):
LGTM

hack/roll_nodes.sh
ettiee: That PR is targeting the 0.20.0 release of kubernetes, should we introduce this change with our 0.20.0 release?

kenju: > That PR is targeting the 0.20.0 release of kubernetes, should we introduce this change with our 0.20.0 release?

Right, maybe it's better to align the version with our release version.

So shall we revisit this PR when we're upgrading to v0.20.0 then?

ettiee: Yes sounds good to me, I think it's cleaner to align client with server versions

errm (approved):
LGTM for 0.20.0

kenju:
:point_right: #228

PR #200, opened 2021-03-04
[cluster/iam] Retire unnecessary AmazonEKSServicePolicy

takanabe:
## Background

According to the [AWS EKS document](https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html,),

> Prior to April 16, 2020, AmazonEKSServicePolicy was also required and the suggested name was eksServiceRole. With the AWSServiceRoleForAmazonEKS service-linked role, that policy is no longer required for clusters created on or after April 16, 2020.

Now, removing `AmazonEKSServicePolicy` from the attached policy and renaming the IAM role used by EKS control plane from service_role to cluster_role makes sense to catch up with the changes in the official document.

## References

* https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html
* https://docs.aws.amazon.com/eks/latest/userguide/using-service-linked-roles-eks.html

### DIFF
--- a/modules/iam/outputs.tf
+++ b/modules/iam/outputs.tf
@@ -1,6 +1,6 @@
 output "config" {
   value = {
-    service_role = aws_iam_role.eks_service_role.name
+    cluster_role = aws_iam_role.eks_cluster_role.name
     node_role    = aws_iam_role.eks_node.name
   }
   depends_on = [
--- a/main.tf
+++ b/main.tf
@@ -9,7 +9,7 @@ module "vpc" {
 module "iam" {
   source = "./modules/iam"
 
-  service_role_name = "eksServiceRole-${var.cluster_name}"
+  cluster_role_name = "eksClusterRole-${var.cluster_name}"
   node_role_name    = "EKSNode-${var.cluster_name}"
 }
 
--- a/modules/cluster/main.tf
+++ b/modules/cluster/main.tf
@@ -2,13 +2,13 @@
   EKS control plane
 */
 
-data "aws_iam_role" "service_role" {
-  name = var.iam_config.service_role
+data "aws_iam_role" "cluster_role" {
+  name = var.iam_config.cluster_role
 }
 
 resource "aws_eks_cluster" "control_plane" {
   name     = var.name
-  role_arn = data.aws_iam_role.service_role.arn
+  role_arn = data.aws_iam_role.cluster_role.arn
   tags     = var.tags
 
   version = var.k8s_version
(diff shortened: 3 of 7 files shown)


modules/iam/outputs.tf
```diff
@@ -1,6 +1,6 @@
 output "config" {
   value = {
-    service_role = aws_iam_role.eks_service_role.name
+    cluster_role = aws_iam_role.eks_cluster_role.name
```
demijonathon, line 3: Should the eks_service_policy be removed at L10, given that it's been removed as a resource ?

demijonathon (changes requested):
One question, but otherwise, LGTM

modules/iam/outputs.tf
takanabe: No, I think we need this line. 

The IAM role for an EKS control plane (a.k.a master or cluster) is still necessary with `AmazonEKSClusterPolicy` but the `AmazonEKSServicePolicy` is no more necessary. For that reason, I'd like to change the name of the IAM role.

demijonathon (approved):
LGTM

takanabe:
@ettiee @dan-slinky-ckpd @aidy 

This might be a breaking change and want you to check as well as release owners 📝

ettiee (commented):
Question: Reading the documentation:
> that policy is no longer required for clusters created on or after April 16, 2020.

suggests that it is required for clusters created before  April 16 2020, suggesting we would need to keep in mind backwards compatibility for clusters that were created before that date.. We have some clusters older than that (other teams) but they are not using the IAM module. Do we need to account for this for any other usecase?

errm:
Do we have a non production cluster created before this date that we could test making this change on to see if it breaks things?

ettiee:
> Do we have a non production cluster created before this date that we could test making this change on to see if it breaks things?

The search / ML staging clusters and ML's experimental clusters are that old, probably ML's experimental one would be a good candidate 👍

modules/iam/outputs.tf
dan-slinky-ckpd: @takanabe Line 10, `aws_iam_role_policy_attachment.eks_service_policy` but you've removed it above.

ettiee:
@takanabe taking a fresh look at this so let me know if you think it's gone stale - seems still useful.
If we change the IAM role for the control plane this forces a recreate on the control plane so this is a breaking change for any users who use the IAM module with the cluster module. Maybe we should provide a variable so that users can override the IAM role name to prevent breaking changes for users who use the IAM module, and add a procedure to include the variable to the upgrade notes?

takanabe:
yeah, seems it's still useful but stale for long time. We should open another PR if we need this because I also don't spend time for this PR.

PR #201, opened 2021-03-05
Include `depends_on` in iam module usage docs

jacksmith15:
I believe it is necessary to specify `depends_on` when using the iam module alongside the cluster module, because internally the cluster module uses a `data` resource to fetch the roles.

Example `plan` output:

> NoSuchEntity: The role with name EKSNode cannot be found

I believe that the `data` source in question here is [this one](https://github.com/cookpad/terraform-aws-eks/blob/b09b004b3f683b60949e23189be11acf3ddf22ac/modules/cluster/main.tf#L61).

An example failed plan is [here](https://github.com/cookpad/global-aws/pull/435#issuecomment-790762218) - which was generated from commit `144fd90d` on that PR. Its worth noting that `node_role` and `service_role` vars are specified on the `iam` module resource - which isn't true of the example in these docs. Could be related?

### Notes

 `iam.config.node_role` clearly depends on the resource being created [here](https://github.com/cookpad/terraform-aws-eks/blob/b09b004b3f683b60949e23189be11acf3ddf22ac/modules/iam/outputs.tf#L4), but this isn't made obvious to the cluster module [here](https://github.com/cookpad/terraform-aws-eks/blob/b09b004b3f683b60949e23189be11acf3ddf22ac/modules/cluster/main.tf#L61) which only views `iam_config` as a var.

PR #202, opened 2021-03-10
[Cluster module] Improve dependency detection (DAG) using ARNs of node/service IAM roles instead of name

takanabe:
# Background

dup: https://github.com/cookpad/terraform-aws-eks/pull/201.

The cluster module is intended to resolve dependencies between the cluster module and [IAM roles passed via a variable](https://github.com/cookpad/terraform-aws-eks/blob/b09b004b3f683b60949e23189be11acf3ddf22ac/modules/cluster/variables.tf#L31-L42).

However, there are possibilities that the cluster module cannot detect the dependencies even we pass new IAM roles passed via a variable.

* https://github.com/cookpad/terraform-aws-eks/pull/201#issue-585541413
* https://github.com/cookpad/terraform-aws-eks/pull/201#issuecomment-793647591

I think this is because the cluster module uses `data` sources to retrieve IAM roles with `name`. If we can pass ARNs of IAM role and eliminate the data blocks for IAM role retrieval, the dependency issue will be solved (obvious for Terraform) and we don't have to use `depends_on` for inter-module dependency declaration.

# Goal

* Pass ARNs of IAM role for both node and control plane instead of names of the IAM role

### DIFF
--- a/modules/cluster/outputs.tf
+++ b/modules/cluster/outputs.tf
@@ -6,7 +6,7 @@ locals {
     vpc_id                = var.vpc_config.vpc_id
     private_subnet_ids    = var.vpc_config.private_subnet_ids
     node_security_group   = aws_eks_cluster.control_plane.vpc_config.0.cluster_security_group_id
-    node_instance_profile = var.iam_config.node_role
+    node_instance_profile = regex("^arn:aws:iam::\\d+:role/*.*/(.+)?", var.iam_config.node_role_arn)[0]
     tags                  = var.tags
     dns_cluster_ip        = local.dns_cluster_ip
     aws_ebs_csi_driver    = var.aws_ebs_csi_driver
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -32,13 +32,11 @@ variable "iam_config" {
   type = object({
     service_role_arn = string
     node_role_arn    = string
-    node_role        = string
   })
 
   default = {
     service_role_arn = ""
     node_role_arn    = ""
-    node_role        = "EKSNode"
   }
 
   description = "The ARN of IAM roles used by the cluster, If you use the included IAM module you can provide it's config output variable."
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -30,16 +30,18 @@ variable "vpc_config" {
 
 variable "iam_config" {
   type = object({
-    service_role = string
-    node_role    = string
+    service_role_arn = string
+    node_role_arn    = string
+    node_role        = string
   })
 
   default = {
-    service_role = "eksServiceRole"
-    node_role    = "EKSNode"
+    service_role_arn = ""
+    node_role_arn    = ""
+    node_role        = "EKSNode"
   }
 
-  description = "The IAM roles used by the cluster, If you use the included IAM module you can provide it's config output variable."
+  description = "The ARN of IAM roles used by the cluster, If you use the included IAM module you can provide it's config output variable."
 }
 
 variable "cluster_autoscaler_iam_permissions_boundary" {
--- a/modules/iam/outputs.tf
+++ b/modules/iam/outputs.tf
@@ -1,8 +1,10 @@
 output "config" {
   value = {
-    service_role = aws_iam_role.eks_service_role.name
-    node_role    = aws_iam_role.eks_node.name
+    service_role_arn = aws_iam_role.eks_service_role.arn
+    node_role        = aws_iam_role.eks_node.name
+    node_role_arn    = aws_iam_role.eks_node.arn
   }
+
   depends_on = [
     aws_iam_role_policy_attachment.eks_worker_node,
     aws_iam_role_policy_attachment.eks_cni,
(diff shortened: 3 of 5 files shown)


ettiee:
Re-running this https://github.com/cookpad/terraform-aws-eks/pull/201#issuecomment-793647591
against your branch resolves the failed plan

errm:
This PR breaks the current behaviour ... that if you create the required roles with the default names then you can create a cluster without explicitly passing anything.

Perhaps we could do something more complex that supports both things

e.g.

```
locals{
  service_role_arn = var.iam_config.service_role_arn == "" ? data.aws_iam_role.service_role.arn : var.iam_config.service_role_arn
}
```

errm:
Thinking about it more... I am probibly OK with changing the behaviour here if it means the module is simpler!

Just make sure we add some documentation to `UPGRADING.md`

modules/cluster/variables.tf
```diff
@@ -30,16 +30,18 @@ variable "vpc_config" {
...
+    node_role_arn    = string
+    node_role        = string
   })
 
   default = {
-    service_role = "eksServiceRole"
-    node_role    = "EKSNode"
+    service_role_arn = ""
+    node_role_arn    = ""
+    node_role        = "EKSNode"
```
errm, line 41: I don't think we should need this any more?

modules/iam/outputs.tf
```diff
@@ -1,8 +1,10 @@
 output "config" {
   value = {
-    service_role = aws_iam_role.eks_service_role.name
-    node_role    = aws_iam_role.eks_node.name
+    service_role_arn = aws_iam_role.eks_service_role.arn
+    node_role        = aws_iam_role.eks_node.name
```
errm, line 4: Can we loose this?

takanabe: this role name is still used by [the instance profile](https://github.com/cookpad/terraform-aws-eks/blob/118ad3076872b052e6cc293ea3be500c840a9b21/modules/cluster/outputs.tf#L9-L10). Otherwise, we still need to retrieve the name using `data` sources. I will consider if I can drop the variable with another way.

takanabe: removed with https://github.com/cookpad/terraform-aws-eks/pull/202/commits/33bef42b4a08e247a17ffde92337c4411835d457

takanabe:
Updated the `UPGRADING.md` https://github.com/cookpad/terraform-aws-eks/pull/202/commits/1d19134187dd1c98fef10f604debfee619417912

takanabe:
I'll have  a look at the following errors 📝 

```
TestTerraformAwsEksCluster 2021-03-12T07:27:10Z logger.go:66: Error: InvalidIamInstanceProfileArn.Malformed: The ARN ‘arn:aws:iam::***:role/EKSNode’ is not valid. The expected format is arn:aws:iam::<account-id>:instance-profile/<instance-profile-name>.
TestTerraformAwsEksCluster 2021-03-12T07:27:10Z logger.go:66: 	status code: 400, request id: 8ee12bdb-9892-4175-97cd-8a1d8ccbcb75
TestTerraformAwsEksCluster 2021-03-12T07:27:10Z logger.go:66: 
TestTerraformAwsEksCluster 2021-03-12T07:27:10Z logger.go:66: 
TestTerraformAwsEksCluster 2021-03-12T07:27:10Z retry.go:80: Returning due to fatal error: FatalError{Underlying: error while running command: exit status 1; 
Error: InvalidIamInstanceProfileArn.Malformed: The ARN ‘arn:aws:iam::***:role/EKSNode’ is not valid. The expected format is arn:aws:iam::<account-id>:instance-profile/<instance-profile-name>.
	status code: 400, request id: 8ee12bdb-9892-4175-97cd-8a1d8ccbcb75

}
```

modules/cluster/outputs.tf
```diff
@@ -6,7 +6,7 @@ locals {
     vpc_id                = var.vpc_config.vpc_id
     private_subnet_ids    = var.vpc_config.private_subnet_ids
     node_security_group   = aws_eks_cluster.control_plane.vpc_config.0.cluster_security_group_id
-    node_instance_profile = var.iam_config.node_role
+    node_instance_profile = regex("^arn:aws:iam::\\d+:role/*.*/(.+)?", var.iam_config.node_role_arn)[0]
```
takanabe, line 9: perhaps this is not always true when we pass `iam_config.node_role_arn` without the IAM module. Let me consider more 🤔

takanabe: should be fine to update with a variable for [`asg_node_group`](https://github.com/cookpad/terraform-aws-eks/blob/7f79fd495912bb03243abd8af83c62b517b6af2a/modules/asg_node_group/variables.tf#L9-L9). Let me keep the change.

takanabe:
rebased to solve conflicts 📝

takanabe:
test this please

takanabe:
seems I need more efforts to pass the CI but I totally forget the details of this implementations. I'll open another PR if we really need this changes. Thank you for reviewing PR.

PR #203, opened 2021-03-11
remove deprecated failure-domain.beta.kubernetes.io/zone label

dan-slinky-ckpd:


modules/asg_node_group/main.tf
```diff
@@ -177,12 +177,6 @@ resource "aws_autoscaling_group" "nodes" {
     propagate_at_launch = false
   }
 
-  tag {
```
mtpereira, line 180: :question: Shouldn't we replace this with the new label, instead of just removing it?

dan-slinky-ckpd: Already in place: https://github.com/cookpad/terraform-aws-eks/blob/remove-deprecated-zone-label/modules/asg_node_group/main.tf#L175

mtpereira (approved):

errm (approved):

PR #204, opened 2021-03-12
Initial 1.19 support upgrade

ettiee:
Ref https://github.com/cookpad/terraform-aws-eks/issues/197

- Upgrade cluster / node group versions
- Upgrade core addons to recommended versions (coredns / kube-proxy)
- Upgrade other addons to latest compatible versions

#### For discussion

[`EKS no longer adds `kubernetes.io/cluster/<cluster-name>` tags to subnets`](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html)
-
We are on the latest load-balancer-controller so we do not need to support these tags for compatibility. However, these tags are required for auto discovery on load balancer creation. I think this is the way we should encourage teams to operate (and is how they do now). The implication for PR changes:
- Add tags to subnets in vpc module
- remove lifecycle change ignore on tags in vpc module
- Recommend users to codify these tags if they are not using the vpc module


ettiee:
@aidy @dan-slinky-ckpd @errm let me know your opinion on 

> EKS no longer adds kubernetes.io/cluster/<cluster-name> tags to subnets

If you agree we should default to keeping these tags on subnets I'll fix the tests and make the tags optional in the VPC module.

modules/vpc/variables.tf
```diff
@@ -11,3 +11,7 @@ variable "cidr_block" {
 variable "availability_zones" {
   description = "The availability zones to create subnets in"
 }
+
+variable "cluster_name" {
```
errm, line 15: We currently (at cookpad) have the pattern of many clusters sharing a VPC, perhaps we should make this an array of cluster names?

errm:
> However, these tags are required for auto discovery on load balancer creation.

Could you explain some more about this? Is there a link to documentation about the situations where it would be needed?

ettiee:
> Could you explain some more about this? Is there a link to documentation about the situations where it would be needed?

Thanks @errm 🙏 it's for the auto discovery for aws load balancer controller https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/subnet_discovery/
In practice - without those subnet tags users will need to specify the tag `alb.ingress.kubernetes.io/subnets` on their ingress definitions.

aidy:
> EKS no longer adds kubernetes.io/cluster/ tags to subnets

I think that makes sense to me.

modules/cluster/addons/aws-load-balancer-controller.yaml
```diff
@@ -26,9 +26,9 @@ metadata:
...
...VnBDeU1BWmpROXlFRStTc2c0ZTlQb0UzUXNXdk1sK0ZJaAovN2RyNUF3SkxudXJFZUNqa3draVZ2cm9LazVSWUFnRm8xRkxmRHJISkNNY0NtdXJ0Sm13QmVuUUZKTTVocTVPCkZ5V2ZHUEZZaFFZamZsZE95OHhHZ1hDSWNDbHVPRll5YU91YkZ0aFQrMW1SSlNlNUlMdlhadTBDZ1lFQXhiaXcKMEpXVDJkU2lrVEx4WWszTTFTM0IzNUNKTFhLY2Faa21NN285RFE4ZVpmbnlHeGRCZVdXTXRMVjkrRFFCeHJYUgo1UW8wU3d3eC96N3R4RDluNmJpWWhNdkZDeDVVTmZ4VDg3KzlhcDR2QSt2b05CM2x5TTVQZlp2U1lQMzJEYTRhCjRlUk1HbGpSVXl1cS80cGZTUzRxdXljbUc1SFQ5di92Ny9PMVRtRUNnWUJpdXZ0Z1gzVTVjWFZzZ2J3cFd5ZVYKT2NMbVo3SmUzNUNVbEhMVHhYV2JxRlpaUjhpS2JhS09HallWdzVkTEpycXYxN2hxOStuaCtPdmNabkUyYWVURApkSElnRnBpeGdJYkIxNzV3WHRyY2R1Rzk5OUxhNFloSW9IbWxsVWhmUkNxMXI3cm9NVjVRc3AveDlXNFV1UHRTCjhWc2xaM3JrS1o1aHhjODJWUktjdlFLQmdDdkRoNCtEckNncEFwcTA5ZXpXWDZKSDhJTzFrRWQ0K05uK3hnL3AKQng1ZTNGR2lFMDZmeFNSRHIvTkh3TXd4WnJzKy8yaDg2N280eHRCQ2M0cGVwbmQ3SG5jMlN6VDh6NUxyZVFHWQp6U3BPUFFYWEw1eHBpYlQ2eFdsK0ZnRFY5YUp0aVJHNjRJVnNaMWpMYjV3RVdyYnQreE5WeFArUk5hT1VYdDF6CnlndkJBb0dBTlJsRUR6SkhtZjJoNW1sN3Qra1kyckxvSjNEWnRDZlNaNlQwaCtxWDFjNmpyWVBtbjJ2QkJqOGwKR1FLQTd5RVNXN1J0SWY5VDB4V2ZEeTFLZVJYWHVwMzZoNkNrd3RVdlZwY1d6MWYwSVBlbjNINWdHSnBoVGNxWgo4bHM1VER5bU5FU2V1WEhFQW10MlIzNDIvelhIekpYVGtKU1A4d3RCUzlMZUhJRmlsUFU9Ci0tLS0tRU5EIFJTQSBQUklWQVRFIEtFWS0tLS0tCg==
```
aidy, line 31: hmm, I'm kinda wary about this...

aidy: It feels wrong to have certificate credentials as part of a generic deploy. Perhaps these should be tf vars or something?

ettiee: Yeah it doesn't feel nice - @dan-slinky-ckpd and I had a chat about this a couple of weeks ago.
It's ended up in here because the current process for including helm charts in the module is to template the chart [here](https://github.com/cookpad/terraform-aws-eks/blob/master/hack/generate_addons.sh#L14) then include the manifests. This chart generates the cert when you template https://github.com/aws/eks-charts/blob/0db3d7d464ddf8f0d7fcae8d5c77d5f7276ab216/stable/aws-load-balancer-controller/templates/webhook.yaml#L1

Some discussion on the project https://github.com/aws/eks-charts/issues/347

I guess there was a design choice in the way that we include helm add-ons (template + commit generated manifests) - easier to debug, don't require users to install helm 🤔

I think the solution would be to add a new command to combine the template+deploy for helm charts and use that [here](https://github.com/cookpad/terraform-aws-eks/blob/master/modules/cluster/addons.tf#L136).
For now though - this cert isn't used anywhere I think, we're terminating SSL at the load balancer - I believe the cert is using for integrating with cert manager, but we're not using cert manager with the load balancer controller.

ettiee: We can avoid the cert generation from the helm templating if we use cert-manager https://github.com/fernando-villalba/aws-load-balancer-controller/commit/668af05a45bace8457e0acce55ad2eec81abbfa5 - I think this is the cleanest solution

ettiee: Discussed in person - will handle in a separate PR from issue https://github.com/cookpad/terraform-aws-eks/issues/205

ettiee:
From yesterday's discussions with @aidy - plan to introduce cert-manager to avoid generating self-signed certs for the aws-load-balancer will be done in #205 
I've implemented the more opinionated VPC / subnet tags in the VPC module example - and upgrade notes for users to add them if they terraform their own VPC.
Remaining review can focus on the 1.19 upgrade + updates to critical addons

aidy (approved):

dan-slinky-ckpd (approved):
LGTM. Gimme an RC :)

PR #209, opened 2021-04-23
bind kube-proxy to 127.0.0.1 when IPV6 is not required

ettiee:
Following the advice from https://github.com/aws/amazon-vpc-cni-k8s/issues/1078#issuecomment-804655593
This fix should stop us hitting the bug around unstable aws-node startup. 
I've made the bindAddress configurable via ipv4/6 usage as binding to 127.0.0.1 will not work for ipv6

### DIFF
--- a/modules/cluster/addons/kube-proxy.yaml
+++ b/modules/cluster/addons/kube-proxy.yaml
@@ -9,7 +9,8 @@ metadata:
 data:
   config: |-
     apiVersion: kubeproxy.config.k8s.io/v1alpha1
-    bindAddress: 0.0.0.0
+    # Advice from https://github.com/aws/amazon-vpc-cni-k8s/issues/1078#issuecomment-804655593
+    bindAddress: 127.0.0.1
     clientConnection:
       acceptContentTypes: ""
       burst: 10


modules/cluster/variables.tf
```diff
@@ -246,3 +246,9 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
   default     = ""
   description = "URI of a custom admin container image"
 }
+
+variable "support_ipv6" {
+  type        = bool
+  default     = false
+  description = "Support requilred for IPV6."
```
kenju, line 253:
```suggestion
  description = "Support required for IPV6."
```

modules/cluster/addons.tf
```diff
@@ -59,7 +59,11 @@ module "kube_proxy" {
   config = local.config
   manifest = templatefile(
     "${path.module}/addons/kube-proxy.yaml",
-    { aws_region = data.aws_region.current.name },
+    {
+      aws_region   = data.aws_region.current.name
+      bind_address = var.support_ipv6 ? "0.0.0.0" : "127.0.0.1"
```
errm, line 64: I am not sure the support_ipv6 == true part of this branch is appropriate...

As I understand the comment here https://github.com/aws/amazon-vpc-cni-k8s/issues/1078#issuecomment-804655593 the idea is that we can bypass some detection for v4 vs v6 if we specify 127.0.0.1...

Since I don't think we are using ipv6 stack at all anywhere I am not sure how to test the behaviour ... I think it would be fine to hardcode this, and make a note in the readme that we don't currently support ipv6

errm (changes requested):

modules/cluster/addons.tf
mtpereira, line 64: :exclamation: I don't think this is a great fix since we are allowing any host to connect under IPv4 as well. Seems like kube-proxy doesn't handle this configuration very gracefully!

Regardless, since we do not use IPv6 ATM and we are not supporting it (i.e., we haven't tested it at all), can we simply drop IPv6 for now? 

```suggestion
      bind_address = "127.0.0.1"
```

errm:
Having read what is going on here https://github.com/kubernetes/kubernetes/pull/91725

I think we are ok to hardcode 127.0.0.1

It seems that the comments about incompatibility with a ipv6 network only apply to a single stack ipv6 setup.

On AWS VPC it is not possible to configure a single stack ipv6 network, you can only configure v4 single stack or dualstack.

ettiee:
@errm @mtpereira ♻️  please 🙏

errm (approved):
LGTM

PR #212, opened 2021-06-02
upgrade aws-ebs-csi-driver to 1.0.0

ettiee:

Issue #186, opened 2021-02-11
Remove kube-bench config override


aidy:
Placeholder task.

When kube-bench 0.6.0 lands, we should remove the config override from #179

ettiee:
kube-bench 0.6.x releases are out now https://github.com/aquasecurity/kube-bench/releases

errm:
:memo: I am working on this issue over here https://github.com/aquasecurity/kube-bench/issues/887

takanabe:
I will start checking the status of this issue 📝

PR #213, opened 2021-06-02
upgrade aws-ebs-csi-controller to 1.0.0

ettiee:
Replaces #212

### DIFF
--- a/modules/cluster/addons/kustomize/overlays/aws-ebs-csi-driver/kustomization.yaml
+++ b/modules/cluster/addons/kustomize/overlays/aws-ebs-csi-driver/kustomization.yaml
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 bases:
-- github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=release-0.10
+- github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=release-1.0
 images:
 - name: k8s.gcr.io/provider-aws/aws-ebs-csi-driver
   newTag: v1.0.0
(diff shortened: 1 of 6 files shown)


modules/cluster/addons/kustomize/overlays/aws-ebs-csi-driver/kustomization.yaml
```diff
@@ -1,10 +1,10 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 bases:
-- github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/alpha/?ref=v0.9.0
+- github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=release-0.10
```
errm, line 4:
```suggestion
- github.com/kubernetes-sigs/aws-ebs-csi-driver/deploy/kubernetes/overlays/stable/?ref=release-1.0
```

??

ettiee: Ah thanks - spuriously placed decimal point! 8cb8814 ..

ettiee:
38a04a9 addresses test failure 
Screenshot 2021-06-03 at 10 01 58

but we should fix the kubebench checks properly  in #186

errm (approved):
LGTM

errm:
Because of the way that we apply the configs here `kubectl apply` it looks like this change will leave behind a number of service accounts that are removed or renamed, perhaps we need to mention them in the upgrade notes.

ettiee:
> Because of the way that we apply the configs here with `kubectl apply` it looks like this change will leave behind a number of service accounts that are removed or renamed, perhaps we need to mention them in the upgrade notes.

Ah thanks good point - I'll add to the notes on the release

Issue #224, opened 2021-08-03
Reconsider addons in cluster module


ettiee:
We want to improve the experience of the cluster addons:

- Users may want to configure addons for their own use-case not use our opinionated 
- helm templating then committing the templated manifests is a manual process, can lead to huge yaml files
- Users might want to upgrade the addons on their own schedule rather than use the version we pin to our EKS release
- [kubectl](https://github.com/cookpad/terraform-aws-eks/tree/master/modules/cluster/kubectl) can have token timeouts against the kubernetes api 

Possible approach after discussion with @errm:

- Remove non-critical addons from the module (discuss with teams and agree addons to remove)
- Use [AWS EKS addons](https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html) for critical addons where supported
- Create kustomize bases for addons we want to keep and manage with [kustomize provider](https://registry.terraform.io/providers/kbst/kustomize/latest/docs)

The expectation after this change is that users will manage addons themselves or consume some other packaging format

mtpereira:
:memo: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon

aidy:
Revisit support for new add-ons, and swap out where applicable.

aidy:
Looks like there's now a [EBS CSI eks addon](https://docs.aws.amazon.com/eks/latest/userguide/managing-ebs-csi.html).

Other than that, we should revisit kube-proxy.

aidy:
> Looks like there's now a EBS CSI eks addon.

Actually, that's still a preview build, so let's hold off on that one for this.

```
        {
            "addonName": "aws-ebs-csi-driver",
            "type": "storage",
            "addonVersions": [
                {
                    "addonVersion": "v1.4.0-eksbuild.preview",
                    "architecture": [
                        "amd64"
                    ],
                    "compatibilities": [
                        {
                            "clusterVersion": "1.21",
                            "platformVersions": [
                                "eks.3+"
                            ],
                            "defaultVersion": true
                        }
                    ]
                }
            ]
        }

```

aidy:
Ah, looks like there's still an issue with kube-proxy in 1.21: https://github.com/aws/containers-roadmap/issues/657#issuecomment-966347660

aidy:
So there's no more we can do here for 1.21

aidy:
`aws eks describe-addon-versions --kubernetes-version 1.21`

[...]

PR #225, opened 2021-08-03
remove load-balancer-controller and ingress tests

ettiee:
Contributes #224
- Remove aws-load-balancer-controller from cluster addons
- Remove validateIngress test


mtpereira:
:heavy_check_mark: Change looks good!

:croissant: Re-running flaky tests.

aidy:
I think we should add some context to the commit messages here - I can see what they're doing, but it's unclear what the motivation is.

aidy:
Suggest that we squash these three commits together with a commit message something like:

```
Remove load-balancer-controller

We've decided to remove non-critical addons from this module.
Rather than try to provide a one-size-fits-all setup, we're changing focus to provide a
minimal setup, and expect that users will use another mechanism for additional configuration. 
```

(`remove load-balancer-controller and ingress tests` is unclear English. Could read as `load-balancer-controller tests and ingress tests` or `ingress tests and load-balancer-controller` - but I think removing the tests here is a facet of removing the load-balancer-controller, rather than explicit change in it's own right)

ettiee:
@aidy sure makes sense - could you check the latest squashed commit

mtpereira (approved):
:rocket:

Issue #210, opened 2021-04-26
flakey test suite


ettiee:
I've seen more flakey test runs recently - need to investigate. A few examples, I initially thought we were hitting another race condition between multiple runs, but I've also seen failures when only one is running. Typically see timeouts waiting for some workload/resource to be ready

aidy:
Closing for now, if this re-occurs we'll reopen and revist.

PR #226, opened 2021-08-04
Increase memory request and limit for cluster-autoscaler addon, add p2.large to gpu_node_group example

ettiee:
Contributes #210

- increase memory/limit for cluster-autoscaler
- add p2.xlarge to example asg_node_group module example `gpu_node_group` 

Investigating CI failures on #225 I had a look at the test cluster, and the cluster-autoscaler pod was repeatedly restarting due to OOM errors.
The particular test failure was that the test workload pods wouldn't start in `validateClusterAutoscaler` (if the cluster-autoscaler pod was OOM then it would not be able to trigger scale-up).

### DIFF
--- a/examples/cluster/gpu_node_group/main.tf
+++ b/examples/cluster/gpu_node_group/main.tf
@@ -15,7 +15,7 @@ module "gpu_nodes" {
   gpu             = true
   instance_family = "gpu"
   instance_size   = "2xlarge"
-  instance_types  = ["p3.2xlarge"]
+  instance_types  = ["p2.xlarge", "p3.2xlarge"]
   zone_awareness  = false
   min_size        = 1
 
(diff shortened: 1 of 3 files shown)


kenju (approved):

ettiee:
CI is failing due to insufficient spot capacity on p2.xlarge

> Launching a new EC2 instance. Status Reason: Could not launch Spot Instances. InsufficientInstanceCapacity - There is no Spot capacity available that matches your request. Launching EC2 instance failed.

This should succeed on a rerun when capacity is back, but another failure case for #210

ettiee:
Added `p2.xlarge` to  example `gpu_node_group` at 5dca2e2260f402ba189a61a7331d9994d8016277 to give more diversity to the ASG when testing an example GPU workload. This should give less chance of CI failing due to no spot availability

mtpereira:
:+1: Change looks good, let's hope for some :green_circle: tests! :crossed_fingers:

ettiee:
We have 🟢  tests 🎉

PR #228, opened 2021-08-06
Deprecate --delete-local-data

mtpereira:
When running `hack/roll_nodes.sh`, I got this warning message:

`Flag --delete-local-data has been deprecated, This option is deprecated
and will be deleted. Use --delete-emptydir-data.`

This change fixes that by using `--delete-emptydir-data` instead.

Reference: https://github.com/kubernetes/kubernetes/pull/95076


kenju (approved):
LGTM

will supersede #199

errm (approved):

PR #229, opened 2021-08-06
Simplify 1.19 upgrade notes

aidy:
In this release we upgrade to aws-load-balancer-controller, but then
remove it.

As both actions happen within a single release, it's less confusing to
just say we removed it.


mtpereira (approved):

errm (approved):
Makes sense!

PR #231, opened 2021-08-06
Switch kubectl to exec credential provider

errm:
Passing the token from aws_eks_cluster_auth data source meant
that the token is generated at plan time - as the tokens
are only valid for 15 minutes this can cause issues when there
is some time elapsed between plan and apply, or if the apply takes
more than 15 minutes e.g. when provisioning a new cluster.

See also this documentation: https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#exec-plugins


mtpereira (approved):
LGTM! Tests were able to setup the cluster, so at least it didn't make it worse for sure! :smile:

PR #234, opened 2021-08-09
Remove nvidia-device-plugin from cluster module #224

ettiee:
Contributes #224

After discussing that we would remove the nvidia-device-plugin from the cluster module - this PR removes the plugin from the cluster module and associated tests.

However, I'd like to get the opinion of other contributors on this:
We are still offering GPU node groups as part of this module, so does removing the `nvidia-device-plugin` will reduce our test coverage? We cannot test workloads using the cluster module with `gpu = true` without this plugin. Is that ok? 
cc @errm @mtpereira @aidy


ettiee:
We discussed this offline, and we decided to continue maintaining the `nvidia-device-plugin` so that we don't lose test coverage on the gpu nodegroup functionality.
A specific example: if a user sets `var.gpu=true` in `asg_node_group`, [this label](https://github.com/cookpad/terraform-aws-eks/blob/f8382583b32a5b1189aac25490d7db3cd792c21e/modules/asg_node_group/main.tf#L22) required for the cluster autoscaler. If we were to introduce a bug in this label value the tests would not pick it up and users of the module may not be able to schedule their workloads to GPU instances

PR #235, opened 2021-08-12
Test against latest terraform version

errm:
This PR updates the terraform we run tests against to the latest.

AFACT the modules should still be compatible back to terraform 0.12

❓ Should we just make a note that we only support the latest terraform version, or should we test against multiple versions?

@mattjw @jnavarro86 @ettiee Do you know what versions of terraform we are currently running internally?

IMO upgrading to 1.0 everywhere seems like a good idea to me, especially as 1.0 gives us some certainty about compatibility and support timelines https://www.terraform.io/docs/language/v1-compatibility-promises.html


mattjw:
global-aws is [0.14.2](https://github.com/cookpad/global-aws/blob/main/terraform/.terraform-version) currently.

I agree adopting 1.0 would be a good thing to do. [Looks](https://www.terraform.io/upgrade-guides/1-0.html) like the pathway from 0.14.x is fairly painless. You have any experience with that @errm?

> Upgrade directly to the latest Terraform v1.0 release and attempt a normal Terraform run. If you encounter any new errors, refer to the v0.15 upgrade guide for resolution steps.

Worth a punt with a local `plan`, at least!

Worth sharing in `#cookpad-global-1` for further discussion. cc @jacksmith15 (SPlaT).

errm:
> I agree adopting 1.0 would be a good thing to do. Looks like the pathway from 0.14.x is fairly painless. You have any experience with that @errm?

Well this PR 😂 , yeah IMO there is nothing much to worry about going from 0.14 ... the significant changes to deal with are the provider requirements changes in 0.12 -> 0.14 after that it's mainly enhancements that I don't think break any existing config.

errm:
📝 The current failure is expected until we upgrade the state for `examples/cluster/environment`

ettiee:
Shall we close this? We have an #267 to track and this PR is quite old predating a lot of changes in the module

ettiee:
We will close this as we haven't worked on it for a while. We will track support / tests against newer terraform versions in #267

PR #237, opened 2021-08-17
Fix flaky tests

errm:
Where we were cleaning up after tests by deleting all resources in a
manifest there could be some issues if the namespace gets deleted before
other resources.

e.g. https://github.com/cookpad/terraform-aws-eks/runs/3339274291?check_suite_focus=true

If we just delete the namespace then k8s will handle cleaning up all the
resources in the namespace.

### DIFF
--- a/test/cluster_test.go
+++ b/test/cluster_test.go
@@ -159,7 +159,7 @@ func validateSecretsBehaviour(t *testing.T, kubeconfig string) {
 	namespace := strings.ToLower(random.UniqueId())
 	kubectlOptions := k8s.NewKubectlOptions("", kubeconfig, namespace)
 	secretManifest := fmt.Sprintf(EXAMPLE_SECRET, namespace, namespace)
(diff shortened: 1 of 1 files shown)


ettiee:
I think one more to change in `validateDNS`?

takanabe (approved):
LGTM if you fix https://github.com/cookpad/terraform-aws-eks/pull/237#issuecomment-900184336

errm:
I don’t think this change is relevant for validateDNS since it doesn’t create or delete a namespace. 🤷‍♂️not sure why we didn’t follow that pattern there…

ettiee:
> I don’t think this change is relevant for validateDNS since it doesn’t create or delete a namespace. 🤷‍♂️not sure why we didn’t follow that pattern there…

Ah you're right 💡 .. wonder if we should change it to follow that pattern instead of using default

ettiee (approved):

PR #240, opened 2021-08-24
Add tfsec PR commenter

errm:
tfsec uses static analysis of your terraform templates to spot potential security issues.

https://tfsec.dev/

### DIFF
--- a/.github/workflows/tfsec_pr_commenter.yml
+++ b/.github/workflows/tfsec_pr_commenter.yml
@@ -1,0 +1,14 @@
+name: tfsec-pr-commenter
+on:
+  pull_request:
+jobs:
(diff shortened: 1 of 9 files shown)


mtpereira:
:clap: Really nice improvement!

:question: I've noticed that `tfsec` did not output anything: https://github.com/cookpad/terraform-aws-eks/runs/3411881464?check_suite_focus=true . Is this expected when there are no violations?

mtpereira:
:memo: This is probably something we should add to release notes:

> To have previous default behaviour you now need to set:
> 
> ```
> endpoint_public_access       = true
> endpoint_public_access_cidrs = ["0.0.0.0/0"]
> ```

PR #247, opened 2021-10-12
remove k8s_version variable

ettiee:
We provided a variable for `k8s_version` in the module. This would allow users to set a different k8s version to the github release they are using, which may cause them to use incorrect configuration for that k8s version and use the wrong version of some addons.

We should make it clear that this isn't configurable and instead users should use the appropriate github release

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,6 +5,8 @@
 * Check the notes for the Kubernetes version you are upgrading to at https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html
 * After upgrading the terraform module, remember to follow the [roll nodes](docs/roll_nodes.md) procedure to roll out upgraded nodes to your cluster.
 
+## 1.19 -> 1.20
+[247](https://github.com/cookpad/terraform-aws-eks/pull/247) 💥 Breaking Change. The `k8s_version` variable has been removed. Use the correct version of the module for the k8s version you want to use.
 
 ## 1.18 -> 1.19
 


ettiee:
> Could you elaborate on the reason why you think so?I think the version is configurable and we don't need this PR.

Sure! Let's say a user wants to use the module to set up a cluster with k8s 1.19. They should use [the terraform cluster module](https://registry.terraform.io/modules/cookpad/eks/aws/latest/submodules/cluster) targeting version 1.19. Let's say that user now wants to install a cluster with k8s 1.17, and they use that same terraform module (version 1.19) but set `k8s_version = 1.17.0`. This will configure the control plane and node template to version 1.17, but some other configuration in the module might cause bugs for k8s 1.17.0. For example the [addons](https://github.com/cookpad/terraform-aws-eks/tree/1.19.0/modules/cluster/addons) will be using the versions recommended for k8s 1.19 and might cause bugs on k8s 1.17. There might be some configuration dropped in a later version that's required in an earlier version of k8s - this might be missed if someone uses a later version of the terraform module and configures k8_version to be for an earlier release.

So two reasons for me to open this PR:
- It's confusing for users to configure the k8s version in two places (as a variable, and by pulling the appropriate version of the terraform module)
- Might cause bugs if users set a different `k8s_version` to the one that the release was tested against

takanabe:
Thank you for the explanation. Now, your intention is clear and I think this changes make sense. So, will we change all supported versions' `k8s_version` variables to local scopes?

Issue #176, opened 2021-02-08
Identify tests that can be run locally


ettiee:
The test suite takes 40+ mins - it would be nice if we could identify anything that can be tested locally to find errors before we run CI.
Candidates:
- validate k8s yaml for cluster-addons

PR #249, opened 2021-10-13
use sts-regional-entpoints=true in eks pod identity webhook

### LINKED ISSUE #160: Use regional endpoints for sts:AssumeRoleWithWebIdentity
## Background

We can use Regional AWS STS endpoints instead of the global endpoint to reduce latency, build in redundancy, and increase session token validity when pods execute `sts:AssumeRoleWithWebIdentity` for IRSA. However, K8s manifests don't have `AWS_STS_REGIONAL_ENDPOINTS=regional` to use the regional endpoints. So, I will add the environment variable to the manifests.

## Target

```
> ag -l eks.amazonaws.com/role-arn
modules/cluster/addons/aws-ebs-csi-driver.yaml
modules/cluster/addons/cluster-autoscaler.yaml
modules/cluster/addons/kustomize/overlays/aws-ebs-csi-driver/eks_iam_for_sa.yaml
modules/cluster/addons/kustomize/overlays/aws-alb-ingress-controller/eks_iam_for_sa.yaml
modules/cluster/addons/helm/cluster-autoscaler.yaml
modules/cluster/addons/aws-alb-ingress-controller.yaml
```

## Related info

* [Managing AWS STS in an AWS Region](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html)

ettiee:
Closes #160 

This PR adds the annotation `eks.amazonaws.com/sts-regional-endpoints: "true"` to addon manifests. This configures the [eks pod identity webhook](https://github.com/aws/amazon-eks-pod-identity-webhook) to inject the environment variable `AWS_STS_REGIONAL_ENDPOINTS` to pods when using IAM roles for service accounts. 

📝  Following the documentation at https://github.com/aws/amazon-eks-pod-identity-webhook#aws_sts_regional_endpoints-injection

It is possible to configure the pod identity webhook to set AWS_STS_REGIONAL_ENDPOINTS=regional by default, but the webhook is configured in the EKS managed service so we cannot do that.

I wonder if we should write a small test to ensure all service accounts have been configured with the annotation  `eks.amazonaws.com/sts-regional-endpoints: "true"`? Or is it not behaviour we'd want to enforce on all addons?


ettiee:
Inadvertently contributed to / started https://github.com/cookpad/terraform-aws-eks/issues/176 as part of #160

test/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+			EksAmazonawsComRoleArn              string `yaml:"eks.amazonaws.com/role-arn"`
+			EksAmazonawsComStsRegionalEndpoints string `yaml:"eks.amazonaws.com/sts-regional-endpoints"`
+		} `yaml:"annotations"`
+		Name string `yaml:"name"`
+	} `yaml:"metadata"`
+}
+
+func TestServiceAccountsUseStsRegionalEndpoint(t *testing.T) {
+	addonsPath := "../modules/cluster/addons/"
+	addonFiles, err := ioutil.ReadDir(addonsPath)
```
takanabe, line 27: `ioutil` package is [deprecated in Go 1.16](https://pkg.go.dev/io/ioutil#ReadDir). I know you specified Go 1.15 in CI but I don't think we have reasons to use the old version.

```suggestion
	addonFiles, err := os.ReadDir(addonsPath)
```

test/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	for _, v := range addonFiles {
+		fileName := v.Name()
+		if filepath.Ext(fileName) == ".yaml" {
+			crds := readManifest(addonsPath + fileName)
```
takanabe, line 34:
```suggestion
			crds, err := readManifest(addonsPath + fileName)
			if err != nil {
			    // do something for error handling
			    // e.g
			    // t.Fatalf("err=%#v\n" , err)
			}
```

test/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+						t.Errorf("Service account %s role arn %s has EksAmazonawsComStsRegionalEndpoints=%s, should be 'true'", v.Metadata.Name, v.Metadata.Annotations.EksAmazonawsComRoleArn, v.Metadata.Annotations.EksAmazonawsComStsRegionalEndpoints)
+
+					}
+				}
+			}
+		}
+	}
+}
+
+func readManifest(filename string) []parsedCRD {
```
takanabe, line 47:
```suggestion
func readManifest(filename string) ([]parsedCRD, error) {
```

test/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+	f, err := os.Open(filename)
+	if err != nil {
+		panic(err)
```
takanabe, line 50:
```suggestion
		return nil, err
```

test/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if crd.Kind == "ServiceAccount" {
+			crds = append(crds, crd)
+		}
+
+	}
+	return crds
```
takanabe, line 68:
```suggestion
	return crds, nil
```

test/local_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	d := yaml.NewDecoder(f)
+	var crds []parsedCRD
+	for {
+		var crd parsedCRD
+		err := d.Decode(&crd)
+		if &crd == nil {
+			continue
+		}
+		if errors.Is(err, io.EOF) {
```
takanabe, line 60: Why don't you have general `if err != nil {}`? This can capture `io.EOF` but cannot catch other errors.

takanabe (commented):
Good job, Ettie!  I left several suggestions on your code so other Gophers could be familiar with your code.

.github/workflows/ci.yml
```diff
@@ -53,6 +53,21 @@ jobs:
...
+      - name: 'Checkout'
+        uses: actions/checkout@main
+      - name: 'Install Go'
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15
+      - name: 'Local Tests'
+        run: | 
+          cd test
+          go install
```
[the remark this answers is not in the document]
takanabe: Also could you leave `Makefile` to automate these steps for local tests?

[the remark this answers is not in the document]
ettiee: yeah sure thing

aidy:
> When using IAM permissions at the pod level, in most cases it is desirable to use the sts regional endpoint not the global endpont.

I think the commit message should have a little more context for why desirable.

test/local_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+type parsedCRD struct {
+	APIVersion string `yaml:"apiVersion"`
+	Kind       string `yaml:"kind"`
+	Metadata   struct {
+		Annotations struct {
```
aidy, line 17: Consider breaking these out into their own types, rather than nesting. More idiomatic.

test/local_test.go
```diff
@@ -0,0 +1,74 @@
...
+			crds, err := readManifest(addonsPath + fileName)
+			if err != nil {
+			    // do something for error handling
+			    // e.g
+			    // t.Fatalf("err=%#v\n" , err)
+			}
+			for _, v := range crds {
+				if v.Kind == "ServiceAccount" {
+					if v.Metadata.Annotations.EksAmazonawsComRoleArn != "" && v.Metadata.Annotations.EksAmazonawsComStsRegionalEndpoints != "true" {
+						t.Errorf("Service account %s role arn %s has EksAmazonawsComStsRegionalEndpoints=%s, should be 'true'", v.Metadata.Name, v.Metadata.Annotations.EksAmazonawsComRoleArn, v.Metadata.Annotations.EksAmazonawsComStsRegionalEndpoints)
```
aidy, line 43: Consider something like
```
if v.Metadata.Annotations.EksAmazonawsComRoleArn != "" {
  assert.True(t, v.Metadata.Annotations.Eks..., fmt.Sprintf("%s should be true, ...))
}
```
Makes it a little more obvious what you're actually testing for, imo.

```

File: test/local_test.go (hunk 1 above)
Comment (ettiee) [reply test/local_test.go]: The project was using 1.15 before https://github.com/cookpad/terraform-aws-eks/blob/main/test/go.mod#L3
Happy to upgrade to 1.17 👍

File: test/local_test.go (hunk 6 above)
Comment (aidy) [reply test/local_test.go:65]: Hm, I think the implementation here is trying to parse every yaml entity as a crd, so errors are expected for the not-crds.

That said, would this condition ever fire? I think the `== nil { continue }` would win.

File: test/local_test.go [hunk 10]
```diff
@@ -0,0 +1,74 @@
+package test
+
+import (
+	"errors"
+	"gopkg.in/yaml.v3"
+	"io"
+	"io/ioutil"
```
aidy, line 7: I think you need to re-goimports this

test/local_test.go
```diff
@@ -0,0 +1,74 @@
...
+		fileName := v.Name()
+		if filepath.Ext(fileName) == ".yaml" {
+			crds, err := readManifest(addonsPath + fileName)
+			if err != nil {
+			    // do something for error handling
+			    // e.g
+			    // t.Fatalf("err=%#v\n" , err)
+			}
+			for _, v := range crds {
+				if v.Kind == "ServiceAccount" {
```
aidy, line 41: I don't think you need to double bag this

https://github.com/cookpad/terraform-aws-eks/pull/249/files#diff-b1a079e8d2b72d05971b0c6d094ba1f8adc1ee7528e4f3ae70da8c4aec63294fR68

test/local_test.go
```diff
@@ -0,0 +1,76 @@
...
+		fileName := v.Name()
+		if filepath.Ext(fileName) == ".yaml" {
+			crds, err := readManifest(addonsPath + fileName)
+			if err != nil {
+				t.Fatalf("err=%#v\n", err)
+			}
+			for _, v := range crds {
+				if v.Kind == "ServiceAccount" {
+					if v.Metadata.Annotations.EksAmazonawsComRoleArn != "" {
+						assert.True(t, v.Metadata.Annotations.EksAmazonawsComStsRegionalEndpoints == "true", fmt.Sprintf("%s should be true", v.Metadata.Annotations.EksAmazonawsComStsRegionalEndpoints))
```
aidy, line 46: I think you could specify the type as bool.

If not, assert.Equal()

test/local_test.go
aidy: By that, I mean, you populated crds with only things that had a Kind of ServiceAccount, I don't think you need to filter here.

test/local_test.go
aidy: Oh, it's a literal string, not a yaml bool.

yeah - `assert.Equal()`

test/local_test.go
ettiee: this condition fires when all elements in the yaml file have been parsed, so if we remove this we get an infinite loop

I think this will cover the points?

```
		if errors.Is(err, io.EOF) {
			break
		}
		if err != nil {
			return nil, err
		}
```

aidy: Why doesn't `if &crd == nil { continue }` win? If we've reached the end of the file, I'd expect that we would not got a crd populated.

aidy: Oh! Because it's *never* nil, because it's a defined variable which you're taking a reference from.

aidy: So maybe...
``
if errors.Is(err, io.EOF) {
    break
}

if crd == parsedCRD{} {
   continue
}
```

I think your `if err != nil { return nil, err }` would not work. Because I *think* you'd get an error if the given yaml token couldn't be unmarshalled into the parsedCRD struct.

File: test/local_test.go [hunk 13]
```diff
@@ -0,0 +1,79 @@
...
+		t.Error(err)
+	}
+	for _, v := range addonFiles {
+		fileName := v.Name()
+		if filepath.Ext(fileName) == ".yaml" {
+			crds, err := readManifest(addonsPath + fileName)
+			if err != nil {
+				t.Fatalf("err=%#v\n", err)
+			}
+			for _, v := range crds {
```
aidy, line 43: Maybe overkill, but I'd be tempted to add a `assert.NotEqual(t, 0, len(crds))`

test/common.go
```diff
@@ -23,7 +23,7 @@ import (
 func deployTerraform(t *testing.T, workingDir string, vars map[string]interface{}) {
 	var terraformOptions *terraform.Options
 
-	if test_structure.IsTestDataPresent(t, test_structure.FormatTestDataPath(workingDir, "TerraformOptions.json")) {
+	if test_structure.IsTestDactaPresent(t, test_structure.FormatTestDataPath(workingDir, "TerraformOptions.json")) {
```
aidy, line 26: Are you deliberately introducing a typo? :)

ettiee: heh no

aidy (approved):

test/local_test.go
ettiee: I pulled this back out - it's a valid usecase to read a manifest and not have any Kind ServiceAccount

Issue #156, opened 2020-10-23
Deprecate use of the root module


errm:
Any serious use of this module should not be using the root module.

Perhaps we should remove any functional parts of the root module?

We can update the example in the readme to show use of cluster + node group modules!

aidy:
Example already updated with #260

PR #252, opened 2021-10-19
Add necessary variable and dependencies for minimum possible config

### LINKED ISSUE #251: Minimum possible config does not work as written
As written, the current setup of this module does not allow the README's minimum possible config:

```
module "eks" {
  source = "cookpad/eks/aws"
  version = "~> 1.16"

  cluster_name       = "hal-9000"
  cidr_block         = "10.4.0.0/16"
  availability_zones = ["us-east-1a", "us-east-1b", "us-east-1c"]
}
```

to be built. 

If the VPC module is not changed --

If I try to create a cluster following the readme:

```
module "eks" {
  source  = "cookpad/eks/aws"
  version = "1.19.0"

  cluster_name = var.cluster_name
  availability_zones = var.availability_zones
  cidr_block = var.vpc_cidr
}

```

and then I run terraform plan, I get this response from terraform:

```
Error: Reference to undeclared input variable

  on .terraform/modules/eks/main.tf line 7, in module "vpc":
   7:   cluster_names      = var.cluster_names

An input variable with the name "cluster_names" has not been declared. Did you
mean "cluster_name"?
```

if I then add cluster_names (cluster_names = [var.cluster_name]) to the module I get this response:

```
Error: Unsupported argument

  on main.tf line 6, in module "eks":
   6:   cluster_names = [var.cluster_name]

An argument named "cluster_names" is not expected here.

```

If the dependency confusion isn't fixed, I get the following when I try to run terraform plan (path is changed because I am running this via my fork):

```
Error: error reading IAM Role (eksServiceRole-noah-berman-lab-cluster): NoSuchEntity: The role with name eksServiceRole-noah-berman-lab-cluster cannot be found.
	status code: 404, request id: foo-1

  on ../../../../terraform-aws-eks/modules/cluster/main.tf line 5, in data "aws_iam_role" "service_role":
   5: data "aws_iam_role" "service_role" {



Error: error reading IAM Role (EKSNode-noah-berman-lab-cluster): NoSuchEntity: The role with name EKSNode-noah-berman-lab-cluster cannot be found.
	status code: 404, request id: foo-2

  on ../../../../terraform-aws-eks/modules/cluster/main.tf line 64, in data "aws_iam_role" "node_role":
  64: data "aws_iam_role" "node_role" {
```

If I want to use this module I need to use each separate sub-module, rather than the minimal config suggested in the readme.

bermannoah:
Closes #251. This PR adds a necessary variable for the VPC module and fixes a dependency issue where resources that need IAM roles would try to be created before the roles they require.

Also -- should I bump the patch or minor version here?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -15,6 +15,9 @@ module "iam" {
 }
 
 module "cluster" {
+  depends_on = [
+    module.iam,
+  ]
   source = "./modules/cluster"
 
   name = var.cluster_name
@@ -29,6 +32,9 @@ module "cluster" {
 }
 
 module "node_group" {
+  depends_on = [
+    module.iam,
+  ]
   source = "./modules/asg_node_group"
 
   cluster_config = module.cluster.config
--- a/variables.tf
+++ b/variables.tf
@@ -3,6 +3,12 @@ variable "cluster_name" {
   description = "A name for the EKS cluster, and the resources it depends on"
 }
 
+variable "cluster_names" {
+  type        = list(string)
+  default     = []
+  description = "Names of the EKS clusters deployed in the VPC."
+}
+
 variable "cidr_block" {
   type        = string
   description = "The CIDR block for the VPC that EKS will run in"
@@ -28,4 +34,3 @@ variable "envelope_encryption_enabled" {
   default     = true
   description = "Should Cluster Envelope Encryption be enabled, if changed after provisioning - forces the cluster to be recreated"
 }
-


ettiee:
Thanks @bermannoah good spot! 
What dependencies are we encoding in the cluster module / vpc module? The cluster module should already have a dependency encoded because [we're passing the config output from the iam module here](https://github.com/cookpad/terraform-aws-eks/blob/main/main.tf#L23). There might be a way we can pass variables to encode the dependencies which feels a bit nicer than putting an explicit module dependency.

We're using terraform 0.12 to test atm so and we support 0.12, I'm not sure if we can use a module dependency in this case because it was [introduced in 0.13](https://www.terraform.io/docs/language/meta-arguments/depends_on.html)

mtpereira:
@ettiee @bermannoah I think what is wrecking the dependency is this, because we point explicitly to the variable: https://github.com/cookpad/terraform-aws-eks/blob/4db0cf3993f6ff194c7f1865a7da4aa8a4df4722/modules/cluster/main.tf#L6

And the variable is defined as something other than the output of the module.

ettiee:
+1 from me for removing root module too. It's not validated in CI and in practicality nobody uses it

ettiee:
Closing this superseded by:
#258 for 1.19 fix
#156 for 1.20

PR #259, opened 2021-11-02
Remove cluster-name tagging from subnets

errm:
* These tags were previously required by aws-load-balancer-controller/alb-ingress-controller
* We no longer ship an ingress controller with the module since #225
* aws-load-balancer-controller doesn't need those tags anyway since v2.1.2 - https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/1773


mtpereira (approved):
:shipit:

ettiee (approved):

PR #260, opened 2021-11-02
Update the README to not include the root module in the main example

errm:
This updates the README to include an example that is inline
with the way we would generally expect the module to be used.

Using the root module means that the deployed cluster/vpc/iam setup
is very inflexible, and can't easily be correctly configured etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,35 +10,55 @@ best practices employed by Cookpad's Global SRE team.
 
 ## Using this module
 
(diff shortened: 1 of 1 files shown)


takanabe (approved):

ettiee (approved):
LGTM
let's merge this now, then we can remove the root module after releasing patches for #251

Issue #242, opened 2021-09-22
`node-group.k8s.cookpad.com/name` can be too big and cause nodes to fail to join the cluster


mtpereira:
## Context

Kubernetes has a [length limitation of 63 characters on labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

In Bottlerocket (which the module uses by default), such a invalid label will cause a node to error out on its configuration step, during boot:

```
[   18.495226] early-boot-config[FAILED] Failed to start Bottlerocket userdata configuration system.

[3074]: See 'systemctl status early-boot-config.service' for details.

Error PATCHing '/settings?tx=bottlerocket-launch': Status 400 when PATCHing /settings?tx=bottlerocket-launch: Json deserialize error: Unable to deserialize into KubernetesLabelValue: Kubernetes label value given invalid input: eks-node-application-engineering-staging-general-purpose-xlarge-spot at line 1 column 1918[

DEPEND] Dependency failed for Applies settings to create config files.
```

## Request

Ensure that this label (and perhaps others we might be using?) that gets passed to the nodes is syntactically correct.

PR #261, opened 2021-11-03
Add validation for k8s label length restrictions

aidy:
connects #242


modules/asg_node_group/versions.tf
```diff
@@ -0,0 +1,10 @@
+terraform {
+  required_version = ">= 0.14"
+
+  required_providers {
+    assert = {
+      source  = "bwoznicki/assert"
+      version = "0.0.1"
+    }
```
ettiee, line 8: aws provider?

ettiee:
📝  upgrading the terraform version requirement to >=0.14 is a breaking change for users using <0.14, and as this is a bug fix we'd want to patch this back to 3 previous minor versions of the module. Looks like you needed to use variable validation for the fix so you had to upgrade to >=0.13. We should consider introducing breaking changes in patch releases

modules/asg_node_group/main.tf
```diff
@@ -17,14 +17,20 @@ locals {
...
-    var.name != "" ? { "node-group.k8s.cookpad.com/name" = var.name } : { "node-group.k8s.cookpad.com/name" = local.name_prefix },
+    { "node-group.k8s.cookpad.com/name" = local.node_group_label },
     var.gpu ? { "nvidia.com/gpu" = "true" } : {},
     var.labels,
   )
 }
 
+data "assert_test" "node_group_label" {
+  test  = length(local.node_group_label) < 64
+  throw = "node-group.k8s.cookpad.com/name label must be 63 characters or less"
```
ettiee, line 31: This error message might not be intuitive to some users - maybe give some indication of what variable to change to not hit the error?
E.g. update name variable
maybe throw contents of node_group_label too so they can inspect

modules/asg_node_group/variables.tf
```diff
@@ -17,6 +17,11 @@ variable "name" {
   type        = string
   default     = ""
   description = "An optional identifier for this node group"
+
+  validation {
```
ettiee, line 21: Maybe covered by https://github.com/cookpad/terraform-aws-eks/pull/261/files#r750337723 already (and vice versa). This validation is more intuitive to a user

ettiee:
> 📝 upgrading the terraform version requirement to >=0.14 is a breaking change for users using <0.14, and as this is a bug fix we'd want to patch this back to 3 previous minor versions of the module. Looks like you needed to use variable validation for the fix so you had to upgrade to >=0.13. We should consider if we want to introduce breaking changes in patch releases

Discussed in person - this isn't really a bug fix to patch back to previous releases. If users hit this they won't have a successfully deployed cluster, so nobody would be patching this into a deployed cluster. It's an improvement we'll include in 1.20.0

modules/asg_node_group/main.tf
aidy: Yeah, I had some conflict about what to do here - the general style of tf errors seems to be to just state the error, rather than suggested resolutions.

modules/asg_node_group/variables.tf
aidy: The vice versa doesn't apply, as if this isn't present, the label is calculated.

Although this is redundant, it seemed more user friendly to duplicate it here. Also, I think the native tf validation fail is nicer.

modules/asg_node_group/versions.tf
aidy: not sure what this means

aidy: Got it.

ettiee (approved):
🚢

Issue #262, opened 2021-11-03
move to setup-terraform github action


ettiee:
https://github.com/hashicorp/terraform-github-actions has been deprecated, superseded by https://github.com/hashicorp/setup-terraform

PR #263, opened 2021-11-03
Switch to setup-terraform github action

aidy:
terraform-github-actions deprecated

connects #262

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,23 +20,15 @@ jobs:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-latest
     steps:
-      - name: 'Checkout'
-        uses: actions/checkout@main
+      - uses: actions/checkout@main
+      - uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_version: 0.12.24
       - name: 'Terraform Format'
-        uses: hashicorp/terraform-github-actions@master
-        with:
-          tf_actions_version: 0.12.24
-          tf_actions_subcommand: 'fmt'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -20,21 +20,17 @@ jobs:
...
-      - name: 'Terraform Format'
-        uses: hashicorp/terraform-github-actions@master
+      - uses: actions/checkout@main
+      - uses: hashicorp/setup-terraform@v1
         with:
-          tf_actions_version: 0.12.24
-          tf_actions_subcommand: 'fmt'
+          terraform_version: 0.12.24
+      - name: 'Terraform Format'
+        run: terraform fmt -check
```
errm, line 28: Might want to add -recursive here

aidy: Ah, yes - good call.

.github/workflows/ci.yml
```diff
@@ -20,23 +20,15 @@ jobs:
...
-      - name: 'Terraform Format'
-        uses: hashicorp/terraform-github-actions@master
+      - uses: actions/checkout@main
+      - uses: hashicorp/setup-terraform@v1
         with:
-          tf_actions_version: 0.12.24
-          tf_actions_subcommand: 'fmt'
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          terraform_version: 0.12.24
```
ettiee, line 26: Spotted that this doesn't line up with https://github.com/cookpad/terraform-aws-eks/blob/main/.terraform-version
Could we improve this by reading that file to an environment variable to use in `terraform_version` ? Would be nice to only define it in 1 place in the project

aidy: Yes - we should do something to have them sync.

Not here though - keeping it as 0.12.24 so it's an atomic change.

PR #272, opened 2021-11-16
require aws provider version >= 3.49.0

ettiee:
Since #240 we require secret encrytion on all clusters. Users hit this bug after enabling secret encryption hashicorp/terraform-provider-aws#19986 unless using aws provider >=3.49.0

NB: for the 1.20 release this will be superseded by #261, but intending to patch this to 1.19.2 for users using https://github.com/cookpad/terraform-aws-eks/pull/240 in 1.19.1 as they will hit 🐛  hashicorp/terraform-provider-aws#19986

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       - uses: actions/checkout@main
       - uses: hashicorp/setup-terraform@v1
         with:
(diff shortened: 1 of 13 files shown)


ettiee:
This will probably be superseded by #261

aidy (approved):

PR #275, opened 2021-11-24
Remove root module

aidy:
We don't intend that people should use the root module, as it's
inflexible for configuration of components.

connects #156


ettiee:
Also remove `variables.tf` and `outputs.tf` ?

aidy:
> Also remove variables.tf and outputs.tf ?

Oh, yeah - I was obviously asleep that day :/

ettiee (approved):
🚢

PR #277, opened 2021-12-01
source path typo

dangen-effy:


ettiee:
@dangen-effy thank you for your PR 😃  
The double slash notation is used when modules are in package sub-directories https://www.terraform.io/docs/language/modules/sources.html#modules-in-package-sub-directories
Testing this change would return an error on running `terraform init`

PR #279, opened 2021-12-02
Remove prometheus node-exporter from cluster add-ons

ettiee:
In #224 we are simplifying the addons provided in this cluster, only providing critical addons to running workloads on the cluster.
Internally we provide prometheus node-exporter via another method, and no teams are using the addon in the module so it will leave a safe upgrade path for everyone after removing.
Users should use another mechanism to manage the resources for prometheus node-exporter


ettiee:
>  You can currently get g4dn.xlarge capacity by not specifying an Availability Zone in your request or choosing us-east-1a, us-east-1c, us-east-1d, us-east-1f. Launching EC2 instance failed."

Could not spin up on-demand nodes for the tests, re-running for now

aidy (approved):

Issue #288, opened 2022-01-28
InvalidParameterException when endpoint_public_access is true endpoint_public_access_cidrs is not set


shimpeko:
## Issue
InvalidParameterException happens at the second execution of terrafrom apply when endpoint_public_access = true and endpoint_public_access_cidrs is not set.

```
│ Error: error updating EKS Cluster (xxx) VPC config: InvalidParameterException: Cluster is already at the desired configuration with endpointPrivateAccess: true , endpointPublicAccess: true, and Public Endpoint Restrictions: [0.0.0.0/0]
│ {
│   RespMetadata: {
│     StatusCode: 400,
│     RequestID: "xxx"
│   },
│   ClusterName: "xxx",
│   Message_: "Cluster is already at the desired configuration with endpointPrivateAccess: true , endpointPublicAccess: true, and Public Endpoint Restrictions: [0.0.0.0/0]"
│ }
│ 
│   with module.eks_cluster_staging.aws_eks_cluster.control_plane,
│   on .terraform/modules/eks_cluster_staging/modules/cluster/main.tf line 12, in resource "aws_eks_cluster" "control_plane":
│   12: resource "aws_eks_cluster" "control_plane" {
│ 
```

## Reproduce

- Create cluster with endpoint_public_access = true and without setting value for endpoint_public_access_cidrs
- Run terraform apply
- Run terraform apply again

aidy:
Closed with #293

PR #289, opened 2022-01-28
Avoid InvalidParameterException

shimpeko:
close https://github.com/cookpad/terraform-aws-eks/issues/288

When endpoint_public_access_cidrs is `[]`, acutual value for
Public Endpoint Restrictions will be `[0.0.0.0/0]`.

But Terraform try to update the "Public Endpoint Restriction"
to [0.0.0.0/0] from secound apply and onward and throws
InvalidParameterException.

It is required to set `null` or explicitly set some value like
`["0.0.0.0/0"]` to void the error.

I think using null is better than `["0.0.0.0/0"]` as it falls to
default of aws_eks_cluster resource.

### DIFF
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -11,7 +11,7 @@ variable "endpoint_public_access" {
 
 variable "endpoint_public_access_cidrs" {
   type    = list(string)
-  default = []
+  default = null
 }
 
 variable "cluster_log_types" {


ettiee:
@shimpeko could you create recreate your PR from a branch on https://github.com/cookpad/terraform-aws-eks? CI needs a secret there to run. Should be fine through @cookpad/sig-infrastructure

shimpeko:
moved to https://github.com/cookpad/terraform-aws-eks/pull/293

PR #297, opened 2022-03-22
Reapply kube-proxy daemonset via EKS addons

aidy:


aidy:
Pretty sure this still doesn't play for our use-case.

Issue #295, opened 2022-03-03
Add gpu support for bottlerocket 


ettiee:
This is enabled in 1.21 by the [aws-k8s-1.21-nvidia variant from bottlerocket](https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#aws-k8s--nvidia-variants). 

📝  make sure you update the docs if working on this issue https://github.com/cookpad/terraform-aws-eks/blob/25ca1db1397441e408b355db7b312dd2ce107af0/modules/asg_node_group/README.md#bottlerocket

NB: [the device plugin is pre-baked into the AMI](https://github.com/weaveworks/eksctl/issues/4760) so users won't need to manage the nvidia plugin daemonset, and we can remove from cluster addons when using bottlerocket.

aidy:
closed with #298

PR #298, opened 2022-03-23
Add support for GPU enabled bottlerocket nodes

aidy:
#295

### DIFF
--- a/modules/asg_node_group/README.md
+++ b/modules/asg_node_group/README.md
@@ -218,7 +218,7 @@ module "bottlerocket_nodes" {
 enable the [AWS EBS CSI driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) by setting `aws_ebs_csi_driver = true` on the cluster module.
 see: https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#csi-plugin
 
-⚠️ Bottlerocket does not yet [support GPU nodes](https://github.com/bottlerocket-os/bottlerocket/issues/769), do not set `gpu = true` when `bottlerocket = true`, as this may result in an invalid configuration!
+⚠️ Bottlerocket now [supports GPU nodes](https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#aws-k8s--nvidia-variants), set `gpu = true` to enable them. Ensure that you set `instance_types` to a GPU instance type.
 
 📝 If you want to get a shell session on your instances via Bottlerocket's SSM agent
 you will need to attach the `arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore` policy
--- a/examples/cluster/bottlerocket_gpu_node_group/cluster.tf
+++ b/examples/cluster/bottlerocket_gpu_node_group/cluster.tf
@@ -1,0 +1,7 @@
+data "terraform_remote_state" "cluster" {
+  backend = "local"
+
+  config = {
+    path = "${path.module}/../terraform.tfstate"
+  }
+}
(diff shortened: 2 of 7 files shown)


modules/asg_node_group/README.md
```diff
@@ -218,7 +218,7 @@ module "bottlerocket_nodes" {
 enable the [AWS EBS CSI driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver) by setting `aws_ebs_csi_driver = true` on the cluster module.
 see: https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#csi-plugin
 
-⚠️ Bottlerocket does not yet [support GPU nodes](https://github.com/bottlerocket-os/bottlerocket/issues/769), do not set `gpu = true` when `bottlerocket = true`, as this may result in an invalid configuration!
+⚠️ Bottlerocket now [supports GPU nodes](https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#aws-k8s--nvidia-variants), set `gpu = true` to enable them. Ensure that you set `instance_types` to a GPU instance type.
```
ettiee, line 221: Maybe add a note about not requiring the nvidia device plugin when using bottlerocket+gpu because the AMIs come with the device plugin?

ettiee: Or update the daemonset in the module [to satisfy](https://github.com/bottlerocket-os/bottlerocket/blob/develop/QUICKSTART-EKS.md#aws-k8s--nvidia-variants):

> If you already have a daemonset for the device plugin in your cluster, you may need to use taints and tolerations to keep it from running on Bottlerocket nodes.

?

aidy: Yeah. Good call - I suspect there's very limited use-cases for people to be running both bottlerocket and not-bottlerocket gpu nodes, but let's cater for that.

aidy: Hmm... I think we're actually set for this, but now I'm not sure where the nvidia-device-plugin is actually supposed to run.

https://github.com/cookpad/terraform-aws-eks/blob/c39fd8600f2a67bb019ed657da5112315dfdbbf1/modules/asg_node_group/main.tf#L24

https://github.com/cookpad/terraform-aws-eks/blob/c39fd8600f2a67bb019ed657da5112315dfdbbf1/modules/cluster/addons/nvidia-device-plugin.yaml#L77-L79

ettiee: Mm I think we're not set for 

> If you already have a daemonset for the device plugin in your cluster, you may need to use taints and tolerations to keep it from running on Bottlerocket nodes.

I think with the current change we'll tag both bottlerocket and non-bottlerocket nodes with `{ "nvidia.com/gpu" = "true" }`
https://github.com/cookpad/terraform-aws-eks/blob/c39fd8600f2a67bb019ed657da5112315dfdbbf1/modules/asg_node_group/main.tf#L24

But we need to tag bottlerocket nodes with something else to prevent the nvidia device plugin scheduling there with the current toleration

aidy: @ettiee isn't that what the `NoSchedule` effect does?

aidy: Oh, I think I get it. It's a toleration for the corresponding taint.

ettiee: yup and we don't want a toleration for the nvidia plugin on bottlerocket gpu nodes (to solve the recommendation from bottlerocket of not scheduling the driver on bottlerocket nodes)

aidy: hm, I suspect the simplest thing to do would be to add a further label to tag it as a bottlerocket node, and re-jig the nodeSelector.

ettiee: 👍 sounds good

aidy: I think that https://github.com/cookpad/terraform-aws-eks/pull/298/files#diff-571d8d76bc35ff080312258572db903f8fcb59a504d6d7491112722634c42a5b should do the trick.

ettiee (approved):

PR #303, opened 2022-04-11
Use `templatefile` function instead of `hashicorp/terraform` provider…

ettiee:
… for templates

The hashicorp/template provider is now deprecated - hashicorp/terraform-provider-template#85.


aidy:
> Closes #303

This closes itself? I mean... you're not *wrong*...

aidy (approved):

PR #306, opened 2022-05-23
feat: change k8s version 1.21 to 1.22

jportasa:


modules/cluster/addons/kube-proxy.yaml
```diff
@@ -101,7 +101,7 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: spec.nodeName
-        image: 602401143452.dkr.ecr.${aws_region}.amazonaws.com/eks/kube-proxy:v1.21.2-eksbuild.2
```
errm, line 104: In theory we should be able to switch to using the `aws_eks_addon` resource for `kube-proxy` now.

We couldn't before as there was an issue that prevented prometheus from scraping metrics from `kube-proxy` - perhaps we should add a test that this could be done...

errm (changes requested):
You will also need to update the addons managed with the `aws_eks_addon` resource

errm:
Looks like there are updates for `aws-node-termination-handler`, I think updating the helm chart version in `hack/generate_addons.sh` should pull in the latest stuff for that.

Latest helm chart version for `cluster-autoscaler` is `9.18.1` ... not sure what changes that will pull in? Might be worth a look now...

errm:
Can we manage the aws-ebs-csi driver via eks addons too? https://docs.aws.amazon.com/eks/latest/userguide/managing-ebs-csi.html

PR #307, opened 2022-05-23
Upgrade to 1.22

jportasa:


modules/cluster/addons.tf
```diff
@@ -25,28 +25,26 @@ module "critical_addons_node_group" {
...
 
 resource "aws_eks_addon" "vpc-cni" {
   cluster_name      = local.config.name
   addon_name        = "vpc-cni"
-  addon_version     = "v1.10.2-eksbuild.1"
+  addon_version     = "v1.11.0-eksbuild.1"
+  resolve_conflicts = "OVERWRITE"
+}
+
+resource "aws_eks_addon" "kube-proxy" {
```
aidy, line 37: Have you verified that [this](https://github.com/cookpad/terraform-aws-eks/commit/dc5394ac5f253eb3e528cb6ad8bd6471bd075c1c) has been resolved?

jportasa: In theory has been fixed in 1.22 https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html but I'll check if the configmap of kube-proxy binding metricsBindAddress is now to 0.0.0.0 instead of 127.0.0.1

jportasa: I have created with terraform apply the example cluster that is in the /example folder of this repo and the bindaddress of the configmap "kube-proxy-config" is now 0.0.0.0, so the kube-proxy addon issue was fixed.

hack/generate_addons.sh
```diff
@@ -18,9 +18,9 @@ kustomize_build() {
   kustomize build $ADDONS_DIR/kustomize/overlays/$1 > $ADDONS_DIR/$1.yaml
 }
 
-helm_template eks aws-node-termination-handler 0.13.3
-helm_template autoscaler cluster-autoscaler 9.10
-helm_template nvdp nvidia-device-plugin 0.10.0
+helm_template eks aws-node-termination-handler 1.15.0
```
errm, line 21: I get:

```
Error: failed to download "eks/aws-node-termination-handler" at version "1.15.0"
```

I think the latest helm chart version is 0.18.4

jportasa: Fixed to 0.18.4

modules/cluster/addons.tf
```diff
@@ -25,28 +25,26 @@ module "critical_addons_node_group" {
...
+  cluster_name      = local.config.name
+  addon_name        = "kube-proxy"
+  addon_version     = "v1.21.2-eksbuild.2"
```
errm, line 40: The docs recommend `1.22.6-eksbuild.1` https://docs.aws.amazon.com/eks/latest/userguide/managing-kube-proxy.html

errm (changes requested):
Can we mange the ebs-csi driver with an eks addon also https://docs.aws.amazon.com/eks/latest/userguide/managing-ebs-csi.html#adding-ebs-csi-eks-add-on

modules/cluster/addons.tf
jportasa: Thx! fixed

modules/asg_node_group/main.tf
```diff
@@ -120,6 +120,13 @@ resource "aws_launch_template" "config" {
     }
   }
 
+  metadata_options {
+    http_endpoint               = "enabled"
+    http_tokens                 = var.nodes_metadata_http_tokens
+    http_put_response_hop_limit = 1
+    instance_metadata_tags      = "disabled"
```
errm, line 127: are we sure nothing depends on this?

errm: or does it default to "disabled" anyway?

modules/asg_node_group/variables.tf
```diff
@@ -184,3 +184,9 @@ variable "bottlerocket_admin_container_source" {
   default     = ""
   description = "URI of a custom admin container image"
 }
+
+variable "nodes_metadata_http_tokens" {
```
errm, line 188: It might be more robust to have this be a bool, then use a ternary to choose one of the two valid strings..

i.e.

```hcl
http_tokens  = var.nodes_metadata_http_tokens ? "required", "optional"
```

errm: If we did that then the name should be something more like `require_instance_metadata_tokens`

errm:
It might be good to extract the IMDSv2 feature to a separate PR, so it would be easier to backport (cherry-pick) to the older versions later on if we need?

modules/asg_node_group/main.tf
jportasa: instance_metadata_tags      = "disabled".  Is the default option
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/launch_template
is a new feature of  Jan 6, 2022 so very likely no cluster uses it, because in the EKS module there is no refference to this variable.
https://aws.amazon.com/about-aws/whats-new/2022/01/instance-tags-amazon-ec2-instance-metadata-service/

WDYT @errm ?

modules/asg_node_group/variables.tf
jportasa: sorted out

modules/asg_node_group/main.tf
errm: OK makes sense... I didn't know it was a very new feature ...

modules/cluster/addons.tf
```diff
@@ -25,28 +25,33 @@ module "critical_addons_node_group" {
...
 
 resource "aws_eks_addon" "coredns" {
   cluster_name      = local.config.name
   addon_name        = "coredns"
-  addon_version     = "v1.8.4-eksbuild.1"
+  addon_version     = "v1.8.7-eksbuild.1"
+  resolve_conflicts = "OVERWRITE"
+}
+
+resource "aws_eks_addon" "ebs-csi" {
```
errm, line 51: Also need to remove the corresponding `module "aws_ebs_csi_driver"` from this file

jportasa: Removed from this PR everything regarding this addon EBS-CSI, i'll add it in different PR

modules/asg_node_group/versions.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.49"
+      version = "~> 4.16"
```
andrassy, line 7: I think there's going to be a deployment challenge here as the old AWS provider constraint was `~> 3.49` and the new one is `~>4.17`. These two constraints are incompatible so we won't be able to upgrade some clusters to this module version and not others within a single project? 😭

andrassy: Might we be able to change the older 1.21-rc module to `>=3.49` so there's at least a pathway? That seems risky too, though, as that's allowing future major version bumps even beyond 4. 😭

modules/cluster/addons.tf
```diff
@@ -25,31 +25,38 @@ module "critical_addons_node_group" {
...
+  addon_version     = "v1.8.7-eksbuild.1"
   resolve_conflicts = "OVERWRITE"
 }
 
+resource "aws_eks_addon" "ebs-csi" {
+  count                    = local.aws_ebs_csi_driver_iam_role_count
+  cluster_name             = local.config.name
+  addon_name               = "aws-ebs-csi-driver"
+  addon_version            = "v1.6.1-eksbuild.1"
+  service_account_role_arn = aws_iam_role.aws_ebs_csi_driver[0].arn
```
errm, line 56:
```suggestion
  service_account_role_arn = local.aws_ebs_csi_driver_iam_role_arn
```

Otherwise this will fail when `var.aws_ebs_csi_driver_iam_role_arn` is set.

modules/cluster/addons.tf
```diff
@@ -95,12 +102,3 @@ module "nvidia_device_plugin" {
   replace  = true
 }
 
-module "aws_ebs_csi_driver" {
-  source = "./kubectl"
-  config = local.config
-  apply  = var.aws_ebs_csi_driver
```
errm, line 101: `var.aws_ebs_csi_driver` sets if this ebs driver should be used, or if we default to the in tree one, either we need to make the addon conditional on this, or remove the ability to choose which driver is used.

modules/asg_node_group/versions.tf
errm: we can add an upper constraint:

Lets do this:

* for 1.22 change the constraint to: `>= 3.49.0, < 5.0.0`
* Add something to release notes saying that for 1.22 we support provider versions 3 and 4 (but 4 is recommended) and that 1.23 will drop support for 3.
* We will have to be careful not to use any features only available in provider v4 in the 1.22 release and should do some manual validation of the rc against both versions of the provider - but keep the automated tests running against v4.

errm: @jportasa - this might be a big / tricky enough topic that it makes sense to extract this provider version stuff to a separate PR - as this would make it easier to cherry-pick a commit onto 1.21 if we decide that that is needed?

errm (changes requested):

modules/cluster/addons.tf
jportasa: changed, thx

modules/cluster/addons.tf
jportasa: fixed.  Use of CSI is optional, but if chosen to use it, should use the EKS addon.

modules/asg_node_group/versions.tf
jportasa: fixed

PR #310, opened 2022-05-26
Remove template provider

### LINKED ISSUE #308: Remove the usage of `template_cloudinit_config`
This is a follow-up from https://github.com/cookpad/terraform-aws-eks/pull/303

After #303, it turns out that this module is still trying to install `hashicorp/template`. As it turns out, `template_cloudinit_config` module is also being provided by `hashicorp/template` provider, and needed to be replaced as well.

I'm opening this issue to track the progress of removal, and hopefully we can include this change into v1.21 before we rolling out the final release

errm:
Fixes #308

My approach here was to remove a feature to make implementation much simpler - so we will need to update this in the UPGRADING document as a breaking change.


sikachu (approved):
Thank you!

PR #313, opened 2022-05-30
Security - add variable to enable or not IMDSv2

jportasa:


modules/asg_node_group/variables.tf
```diff
@@ -178,3 +178,9 @@ variable "bottlerocket_admin_container_source" {
   default     = ""
   description = "URI of a custom admin container image"
 }
+
+variable "imdsv2_enabled" {
```
errm, line 182: I think name like "required" or "enforced" would be more descriptive

jportasa: changed to imdsv2_required

jportasa: changed to imdsv2_required

piotr-zuralski (approved):

examples/cluster/bottlerocket_gpu_node_group/main.tf
```diff
@@ -24,6 +24,9 @@ module "node_group" {
   instance_types = ["g4dn.xlarge"]
   min_size       = 1
 
+  imdsv2_required = true
```
errm, line 27: do we need to set this everywhere in the examples - it is the default?

jportasa: the default is true, I'll remove it from here just to make it cleaner

PR #318, opened 2022-06-23
Disable install k8s yaml manifests by terraform (can be done by Flux)

jportasa:


errm:
all these addons allready have variables to disable them anyway ... I think we should be concentrating on https://github.com/cookpad/terraform-aws-eks/issues/224 instead

PR #324, opened 2022-08-12
Upgrade k8s 1.23

errm:


hack/generate_addons.sh
```diff
@@ -11,4 +11,4 @@ helm_template() {
   helm template --no-hooks --namespace=kube-system --version $3 -f $ADDONS_DIR/helm/$2.yaml $2 $1/$2${4:-} | grep -v Helm > $ADDONS_DIR/$2.yaml
 }
 
-helm_template autoscaler cluster-autoscaler 9.18.1
+helm_template autoscaler cluster-autoscaler 9.19.3
```
jportasa, line 14: I would remove autoscaler from the module. why? because Flux can install it and with POD priority we can say to the scheduler to schedule 1st PODS with high priority. Other reason is because maybe we want to install other node autoscaler system.

modules/cluster/addons/cluster-autoscaler.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/instance: "cluster-autoscaler"
     app.kubernetes.io/name: "aws-cluster-autoscaler"
-    helm.sh/chart: "cluster-autoscaler-9.18.1"
+    helm.sh/chart: "cluster-autoscaler-9.19.3"
```
jportasa, line 10: same, I would remove autoscaler instlation

hack/generate_addons.sh
jportasa: In case of using autoscaler, we will want to set the param that sets the avg cpu/mem for each node, before scale, staging for example nodes are underutilitzated

errm: Yeah this PR is just to get the new version of k8s up and running with the existing setup.

I am working on some additional changes, including depreciation of the autoscaler addons. There is a quite a bit of complexity here in order to make the module testable! So I am going to make those changes in an additional PR

jportasa: ok! just for curiosity, in autoscaler you have this param:

The scaleDownUtilizationThreshold defines the proportion between requested resources and capacity, which under the value cluster-autoscaler will trigger the scaling down action.

Our default value is 65%, which means in order to scale down, one of the nodes has to have less utilization (CPU/memory) than this threshold.

jportasa (approved):

piotr-zuralski (approved):

PR #327, opened 2022-09-14
Add imdsv2_required flag for critical addons asg

jnavarro86:
The ML team depends on the flag `imdsv2_required` in one of our clusters because we still use `kube2iam` for an app.

We need to be able to pass this flag to the `critical_addons` ASG because `kube2iam` pod can't startup in those nodes. 

This is my first PR into the module. Happy to help with the release process but will need some guidance :)


jnavarro86:
I end up removing the toleration of `kube2iam` daemonset for critical addons

PR #330, opened 2023-02-10
Support migration to self-managed CoreDNS

jacksmith15:
This PR aims to enable users of the module to migrate to self-managed CoreDNS (as opposed to using the EKS addon).

The [included docs](https://github.com/cookpad/terraform-aws-eks/blob/js/search-3950/support-migration-to-self-managed-coredns/docs/self_managed_coredns.md#self-managed-coredns) describe the intended process for users, as well as reasons why a user may want to manage CoreDNS themselves.

tl;dr I would like to be able to scale CoreDNS in my cluster, but am not currently able to due to the way the EKS addon will overwrite any modifications I make.

### DIFF
--- a/modules/cluster/addons.tf
+++ b/modules/cluster/addons.tf
@@ -46,10 +46,12 @@ resource "aws_eks_addon" "kube-proxy" {
 }
 
 resource "aws_eks_addon" "coredns" {
+  count             = var.critical_addons_coredns_enabled == true ? 1 : 0
   cluster_name      = local.config.name
   addon_name        = "coredns"
   addon_version     = "v1.8.7-eksbuild.2"
   resolve_conflicts = "OVERWRITE"
+  preserve          = var.critical_addons_coredns_preserve
   depends_on = [
     module.critical_addons_node_group
   ]
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -196,8 +196,8 @@ variable "critical_addons_coredns_preserve" {
 }
 
 variable "critical_addons_coredns_enabled" {
-  type = string
-  default = true
+  type        = string
+  default     = true
   description = "Whether to enable CoreDNS addon. Recommended for cluster creation, but can be disabled to migrate to self-managed CoreDNS."
 }
 
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -189,6 +189,18 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
   description = "URI of a custom admin container image"
 }
 
+variable "critical_addons_coredns_preserve" {
+  type        = string
+  default     = true
+  description = "Whether to preserve CoreDNS resources when deleting the addon. This can be used to migrate to self-managed CoreDNS."
+}
+
+variable "critical_addons_coredns_enabled" {
+  type = string
+  default = true
+  description = "Whether to enable CoreDNS addon. Recommended for cluster creation, but can be disabled to migrate to self-managed CoreDNS."
+}
+
 variable "security_group_ids" {
   type        = list(string)
   default     = []
--- a/docs/self_managed_coredns.md
+++ b/docs/self_managed_coredns.md
@@ -1,0 +1,24 @@
+# Self-managed CoreDNS
+
+This document describes how to migrate from CoreDNS as an EKS addon to self-managed CoreDNS.
+
+> :warning: It is recommended that new clusters use the EKS addon, even if you intend to immediately migrate to self-managed.
+
+## Steps
+
+1. Ensure that the `critical_addons_coredns_preserve` variable is configured to `true`. This is the default value, and ensures no downtime for DNS during the migration.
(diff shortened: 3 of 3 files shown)


modules/cluster/variables.tf
```diff
@@ -189,6 +189,18 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
   description = "URI of a custom admin container image"
 }
 
+variable "critical_addons_coredns_preserve" {
+  type        = string
+  default     = true
```
jacksmith15, line 194: I think its safe to make this `true` by default. It reduces the number of steps required to migrate, and has no effect for users who do not want to migrate.

errm (approved):
LGTM ...

Please `terraform fmt`

modules/cluster/addons.tf
```diff
@@ -46,10 +46,12 @@ resource "aws_eks_addon" "kube-proxy" {
 }
 
 resource "aws_eks_addon" "coredns" {
+  count             = var.critical_addons_coredns_enabled == true ? 1 : 0
```
errm, line 49: I think this is risky actually ... please test, but I think it may force a delete and recreate on an existing cluster!

errm (changes requested):

modules/cluster/addons.tf
errm: I am guessing we will need some instructions in UPGRADING to explain the manual state migration steps that will be needed!

jacksmith15: Could I request some support on (or information about) the testing process?

The [docs mention pre-production testing](https://github.com/cookpad/terraform-aws-eks/blob/main/RELEASING.md#branching) - do we have a test cluster configured for this purpose?

jacksmith15:
> question Did you ask AWS if they can make changes to the addon to support the behaviour you need!

Short answer: no

Long answer:

They do _already_ support the behaviour I need, but only in a roundabout manner. The EKS addon accepts a `configuration_values` parameter - which has the following parameters:

- `corefile` (configmap contents)
- `nodeSelector`
- `replicaCount`
- `resources`
- `computeType`

<details><summary><b>Configuration values schema</b></summary>

```json
{
  "$ref": "#/definitions/Coredns",
  "$schema": "http://json-schema.org/draft-06/schema#",
  "definitions": {
    "Coredns": {
      "additionalProperties": false,
      "properties": {
        "computeType": {
          "type": "string"
        },
        "corefile": {
          "description": "Entire corefile contents to use with installation",
          "type": "string"
        },
        "nodeSelector": {
          "additionalProperties": {
            "type": "string"
          },
          "type": "object"
        },
        "replicaCount": {
          "type": "integer"
        },
        "resources": {
          "$ref": "#/definitions/Resources"
        }
      },
      "title": "Coredns",
      "type": "object"
    },
    "Limits": {
      "additionalProperties": false,
      "properties": {
        "cpu": {
          "type": "string"
        },
        "memory": {
          "type": "string"
        }
      },
      "title": "Limits",
      "type": "object"
    },
    "Resources": {
      "additionalProperties": false,
      "properties": {
        "limits": {
          "$ref": "#/definitions/Limits"
        },
        "requests": {
          "$ref": "#/definitions/Limits"
        }
      },
      "title": "Resources",
      "type": "object"
    }
  }
}
```

</details>

In theory the `replicaCount` and `resources` fields are sufficient for my particular purposes. The reason I proposed the current approach is because my goal is to be able to easily respond to an outage cause by unusually high DNS load. In that case, I believe the time-to-recover will be lower if the process for tweaking vertical and horizontal scale of CoreDNS is familiar to the responder.

This is a loosely held opinion though - I'd be happy to instead expose the `configuration_values` option in this Terraform module if you think its more appropriate.

jacksmith15:
Thanks for prompt review btw!

ettiee:
> In theory the replicaCount and resources fields are sufficient for my particular purposes. The reason I proposed the current approach is because my goal is to be able to easily respond to an outage cause by unusually high DNS load. In that case, I believe the time-to-recover will be lower if the process for tweaking vertical and horizontal scale of CoreDNS is familiar to the responder (i.e. modify a k8s manifest).

> This is a loosely held opinion though - I'd be happy to instead expose the configuration_values option in this Terraform module if you think its more appropriate.

I'm considering exposing `configuration_values` for the EBS CSI addon, so if we make this a standard pattern it may become more familiar to the teams. And still the added benefit of not having to manage a kubernetes manifest / helm release for things provided as EKS addons

jacksmith15:
I think it probably makes sense to expose the configuration values in this case :+1: 

I'll prepare a PR which exposes this for each of the relevant addons

ettiee:
> I think it probably makes sense to expose the configuration values in this case 👍
> 
> I'll prepare a PR which exposes this for each of the relevant addons

Sounds good! I'll probably change my approach then and go forward with your change.
you'll need to upgrade the addon to
addon_version = "v1.10.0-eksbuild.1" 
for EBS CSI driver to support this, and some version for the other addons too I expect. 
When you've got your PR ready we can cherry pick onto a release for 1.22

PR #331, opened 2023-02-10
Support custom configuration values for addons

jacksmith15:
This PR exposes addon [`configuration_values`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_addon#configuration_values) in the cluster module's variables.

This will allow users to configure addons (for example, increasing no. of CoreDNS replicas).

This change can also be cherry-picked onto the 1.22 release branch, but we will need to upgrade the `aws-ebs-csi-driver` addon on that branch to `v1.10.0-eksbuild.1` to support configuration values.

### DIFF
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -190,25 +190,72 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
 }
 
 variable "critical_addons_vpc-cni_configuration_values" {
-  type        = string
+  type = object(
+    {
+      resources = object({ limits = object({ cpu = string, memory = string }), requests = object({ cpu = string, memory = string }) }),
+      env = object(
+        {
+          ADDITIONAL_ENI_TAGS                   = string,
+          ANNOTATE_POD_IP                       = string,
+          AWS_VPC_CNI_NODE_PORT_SUPPORT         = string,
+          AWS_VPC_ENI_MTU                       = string,
+          AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER    = string,
+          AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG    = string,
+          AWS_VPC_K8S_CNI_EXTERNALSNAT          = string,
+          AWS_VPC_K8S_CNI_LOGLEVEL              = string,
+          AWS_VPC_K8S_CNI_LOG_FILE              = string,
+          AWS_VPC_K8S_CNI_RANDOMIZESNAT         = string,
+          AWS_VPC_K8S_CNI_VETHPREFIX            = string,
+          AWS_VPC_K8S_PLUGIN_LOG_FILE           = string,
+          AWS_VPC_K8S_PLUGIN_LOG_LEVEL          = string,
+          DISABLE_INTROSPECTION                 = string,
+          DISABLE_METRICS                       = string,
+          DISABLE_NETWORK_RESOURCE_PROVISIONING = string,
+          ENABLE_POD_ENI                        = string,
+          ENABLE_PREFIX_DELEGATION              = string,
+          WARM_ENI_TARGET                       = string,
+          WARM_PREFIX_TARGET                    = string
+        }
+      )
+    }
+  )
   default     = null
   description = "Configuration values passed to the vpc-cni EKS addon."
 }
 
 variable "critical_addons_kube-proxy_configuration_values" {
-  type        = string
+  type = object(
+    {
+      ipvs      = object({ scheduler = string }),
+      mode      = string, # "iptables" or "ipvs"
+      resources = object({ limits = object({ cpu = string, memory = string }), requests = object({ cpu = string, memory = string }) })
+    }
+  )
   default     = null
   description = "Configuration values passed to the kube-proxy EKS addon."
 }
 
 variable "critical_addons_coredns_configuration_values" {
-  type        = string
+  type = object(
+    {
+      computeType  = string,
+      corefile     = string,
+      nodeSelector = map(string),
+      replicaCount = number,
+      resources    = object({ limits = object({ cpu = string, memory = string }), requests = object({ cpu = string, memory = string }) })
+    }
+  )
   default     = null
   description = "Configuration values passed to the coredns EKS addon."
 }
 
 variable "critical_addons_ebs-csi_configuration_values" {
-  type        = string
+  type = object(
+    {
+      controller = object({ extraVolumeTags = map(string) }),
+      node       = object({ tolerateAllTaints = bool, volumeAttachLimit = number })
+    }
+  )
   default     = null
   description = "Configuration values passed to the ebs-csi EKS addon."
 }
--- a/modules/cluster/variables.tf
+++ b/modules/cluster/variables.tf
@@ -190,72 +190,25 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
 }
 
 variable "critical_addons_vpc-cni_configuration_values" {
-  type = object(
-    {
-      resources = object({ limits = object({ cpu = string, memory = string }), requests = object({ cpu = string, memory = string }) }),
-      env = object(
-        {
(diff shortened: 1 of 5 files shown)


modules/cluster/variables.tf
```diff
@@ -189,6 +189,30 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
...
+
+variable "critical_addons_coredns_configuration_values" {
+  type        = string
+  default     = null
+  description = "Configuration values passed to the coredns EKS addon."
+}
+
+variable "critical_addons_ebs-csi_configuration_values" {
+  type        = string
+  default     = null
```
jacksmith15, line 212: I had to double-check, but passing `null` through to the `eks_addon` resource is equivalent to not passing the option

https://developer.hashicorp.com/terraform/language/expressions/types#null

errm:
Please update https://github.com/cookpad/terraform-aws-eks/blob/main/modules/cluster/providers.tf#L5 so that we require aws provider `>= 4.47.0`

modules/cluster/variables.tf
```diff
@@ -189,6 +189,30 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
   description = "URI of a custom admin container image"
 }
 
+variable "critical_addons_vpc-cni_configuration_values" {
```
jacksmith15, line 192: Possibly contentious variable naming - I've preserved the addon name as-is, including the kebab case. _Possibly_ makes it easier to identify that `vpc-cni` is the addon, and `configuration_values` is the argument.

Happy to go full snake case.

jacksmith15:
:eyes: Investigating the failed dependency resolution

modules/cluster/variables.tf
errm: On the whole our rule is to just snake case everything for consistency ... 🤷 I am 50/50 on this one though!

modules/cluster/variables.tf
```diff
@@ -189,6 +189,30 @@ variable "critical_addons_node_group_bottlerocket_admin_container_source" {
...
+
+variable "critical_addons_kube-proxy_configuration_values" {
+  type        = string
+  default     = null
+  description = "Configuration values passed to the kube-proxy EKS addon."
+}
+
+variable "critical_addons_coredns_configuration_values" {
+  type        = string
+  default     = null
```
errm, line 206: We could use structural typing here to do some kind of validation on the inputs... unless this is already handled internally by the provider???

```suggestion
  type = object({ computeType = string, corefile=string, nodeSelector=object(string), replicaCount=number, resources=object(string)})
  default  = null
```

jacksmith15: Different versions of an EKS addon have different schema for their configuration values - so the provider simply transparently passes the json-encoded string, and leaves mapping the addon version <-> configuration value schema to the user.

Because we hard-code the addon versions, it _is_ technically possible for us to provide a well typed interface with the configuration values. I will do this.

I should note a drawback that this adds more steps to our upgrade process for this module. We will need through each addon version we've updated, fetch the latest configuration values schema, and map that into the variables.

We should probably do this activity anyway however, to notify users of breaking changes in the configuration value schema.


> :memo: Note to self: command to check configuration value schema is
> ```
> $ envchain aws aws --profile=cookpad-global-1 eks describe-addon-configuration --addon-name aws-ebs-csi-driver --addon-version v1.10.0-eksbuild.1 | jq '.configurationSchema' -r | jq
> ```

ettiee:
We can update [this example](https://github.com/cookpad/terraform-aws-eks/blob/main/examples/cluster/main.tf#L31) to include some sensible addon configuration values for each addon to include the functionality in the test suite. I don't think we need to test each configuration path just that it gets parsed correctly

modules/cluster/variables.tf
ettiee: CI is failing with 

> Message_: "ConfigurationValue provided in request is not supported: Json schema validation failed with error: [$: null found, object expected]"

so we need to check this. Checking the jsonschemas for the addon configurations we can pass an empty object as the default

modules/cluster/variables.tf
ettiee: If we do this I think we could look at adding a unit test to check our well typed interface against the jsonschema for the addon

ettiee: I think we can use https://pkg.go.dev/github.com/hashicorp/terraform-config-inspect/tfconfig#LoadModule within a unit test to load [the cluster module and its Variables](https://pkg.go.dev/github.com/hashicorp/terraform-config-inspect/tfconfig#Module)

modules/cluster/variables.tf
ettiee: or keep null as default and don't call `jsonencode` in the `aws_eks_addon` resources

jacksmith15: Ah of course - the `jsonencode` is the source of the problem.

Its not _totally_ clear what the expected behaviour is for `configuration_values = "{}"`, because whilst none of the fields are included in `required`, they also have no defined jsonschema `default`.

ettiee: Ah yes you're right - perhaps better to avoid the call to `jsonencode` then 👍

jacksmith15:
I am somewhat undecided on the test strategy.

I think its most important we test the case where no `configuration_values` are passed by the user, as this is the most common use case, and also ensures backwards-compatibility.

However I note that we appear to have only one test cluster configured - perhaps this is to keep test times short?

Two options:
1. As a compromise, we could provide configuration values for some addons, but not for others. This could be a little flaky, but would partly test both cases.
2. We could add another cluster module invocation to the test suite, to explicitly test each addon with a set of configuration values

ettiee:
> I think its most important we test the case where no configuration_values are passed by the user, as this is the most common use case, and also ensures backwards-compatibility.

Yes this makes sense 👍 

> However I note that we appear to have only one test cluster configured - perhaps this is to keep test times short?

Yes the integration tests take a while to run,  however [these tests](https://github.com/cookpad/terraform-aws-eks/blob/main/test/local_test.go) run locally so if you want to test more permutations of configuration values you could write some local tests as I mentioned in https://github.com/cookpad/terraform-aws-eks/pull/331#discussion_r1104365072? This would allow you to explicitly test each configuration for validity with the corresponding jsonschema. Then maybe we could go with your suggestion for the integration tests for a partial test of both cases:

> As a compromise, we could provide configuration values for some addons, but not for others. This could be a little flaky, but would partly test both cases.

jacksmith15:
From investigation I think that the current approach of structural typing will not work, because:

- An HCL structural type `object({ foo=string, bar=string })` will always require all defined properties to be passed
- The default configuration values for EKS addons are not documented _anywhere_, as far as I can see (the [related docs](https://docs.aws.amazon.com/eks/latest/userguide/eks-add-ons.html) only mention how to find the schema, which includes no defaults)
- The recommended use of configuration values is to only override the fields the user wishes to override (see the [Creating an addon docs](https://docs.aws.amazon.com/eks/latest/userguide/managing-add-ons.html)).

Using the structural typing would require users to provide all possible configuration values when there is no clear way to identify the default values.

So for example, if I want to set my coredns replicas to 3, I would need to figure out valid configuration values for `nodeSelector`, `resources`, `computeType` and `Corefile`, without being able to view the current default values. This is especially a problem for `vpc-cni`, which has 21 environment variable fields to configure.

As such I will switch back to passing configuration values as strings.

We may still want to run some tests against some configuration values when we upgrade the addon version, as this will allow us to document any breaking changes to the `configuration_values` schema. Notably I don't see anywhere that AWS would document such breaking changes though...

modules/cluster/variables.tf
jacksmith15: I chose a different direction - see https://github.com/cookpad/terraform-aws-eks/pull/331#issuecomment-1428155603

ettiee (approved):

PR #337, opened 2023-03-17
Enhancements from web-platform

errm:
📝 These are some enhancements we have been using on the web platform clusters for some time!

It would be great to merge them in so we can use the mainline release!

* Removes the provisioner that "waits" for the cluster to be ready, and instead adding a similar check to the kubectl command fixes a lot of errors that can require a re-run when provisioning a cluster.
* In our environment we need to give all our IAM roles a specific prefix - this change (optionally) allows this to be done!

### DIFF
--- a/modules/cluster/aws_ebs_csi_driver_iam.tf
+++ b/modules/cluster/aws_ebs_csi_driver_iam.tf
@@ -23,7 +23,7 @@ data "aws_iam_policy_document" "aws_ebs_csi_driver_assume_role_policy" {
 
 resource "aws_iam_role" "aws_ebs_csi_driver" {
   count                = local.aws_ebs_csi_driver_iam_role_count
-  name                 = "EksEBSCSIDriver-${var.name}"
+  name                 = "${var.iam_role_name_prefix}EksEBSCSIDriver-${var.name}"
   assume_role_policy   = data.aws_iam_policy_document.aws_ebs_csi_driver_assume_role_policy.json
   permissions_boundary = var.aws_ebs_csi_driver_iam_permissions_boundary
 }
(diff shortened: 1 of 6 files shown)


modules/cluster/aws_ebs_csi_driver_iam.tf
```diff
@@ -23,7 +23,7 @@ data "aws_iam_policy_document" "aws_ebs_csi_driver_assume_role_policy" {
 
 resource "aws_iam_role" "aws_ebs_csi_driver" {
   count                = local.aws_ebs_csi_driver_iam_role_count
-  name                 = "EksEBSCSIDriver-${var.name}"
+  name                 = "${var.iam_role_name_prefix}EksEBSCSIDriver-${var.name}"
```
ettiee, line 26:
```suggestion
  name                 = "${var.iam_role_name_prefix}-EksEBSCSIDriver-${var.name}"
```
would be nice for readability - or do you need this as-written to be compatible with with what you've implemented for the web platform clusters?

pray: I think this makes the prefix non optional (default role name will be `-EksEBSCSIDriver-nanana`)?

errm: Perhaps we can do something like
```suggestion
  name                 = join("-", compact([var.iam_role_name_prefix, "EksEBSCSIDriver", var.name]))
```

ettiee: ah yes neater than the suggestion I was mid-writing with a var null condition

PR #340, opened 2023-04-28
Tags propagation (respect default_tags, node group ASG, critical node group via cluster)

sorah:
This PR improves resource tagging for cost allocation by:

- respect AWS provider's default_tags. ASG needs special handling using `aws_default_tags` data provider.
  - https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/default_tags
- Inherit tags to all possible resources defined through launch templates
  - https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RequestLaunchTemplateData.html
  - https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchTemplateTagSpecificationRequest.html
- critical_addons_node_group does not respect `var.tags` specified for cluster module itself

### DIFF
--- a/modules/asg_node_group/main.tf
+++ b/modules/asg_node_group/main.tf
@@ -1,3 +1,5 @@
+data "aws_default_tags" "current" {}
+
 locals {
   k8s_version = "1.23"
   preset_instance_families = {
@@ -21,7 +23,7 @@ locals {
   root_device_mappings = reverse(tolist(data.aws_ami.image.block_device_mappings))[0]
   autoscaler_tags      = var.cluster_autoscaler ? { "k8s.io/cluster-autoscaler/enabled" = "true", "k8s.io/cluster-autoscaler/${var.cluster_config.name}" = "owned" } : {}
   bottlerocket_tags    = var.bottlerocket ? { "Name" = "eks-node-${var.cluster_config.name}" } : {}
-  tags                 = merge(var.cluster_config.tags, var.tags, { "kubernetes.io/cluster/${var.cluster_config.name}" = "owned" }, local.autoscaler_tags, local.bottlerocket_tags)
+  tags                 = merge(data.aws_default_tags.current.tags, var.cluster_config.tags, var.tags, { "kubernetes.io/cluster/${var.cluster_config.name}" = "owned" }, local.autoscaler_tags, local.bottlerocket_tags)
   node_group_label     = var.name != "" ? var.name : local.name_prefix
   cloud_config = templatefile(
     "${path.module}/cloud_config.tpl",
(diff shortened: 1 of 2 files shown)


sorah:
@cookpad/platform-task-force ptal

jportasa:
For the inminent 1.25 release that  @errm  is preparing, and we will deploy soon, the launch templates will change a lot because we are going to use:
- Fargate for critical nodes (where we will place critical k8s services like hube-system, flux, karpenter,...)
- Karpenter for the other nodes, where karpenter will setup the tags, not terraform (the idea is to use k8s as much as posible to setup everything).
- Addons will disappear in terraform, and will be installed by Flux in k8s.

Therefore the EKS TF module will be very simplified, so I don't know if worth changing the tags with this config.

sorah:
> * Karpenter for the other nodes, where karpenter will setup the tags, not terraform (the idea is to use k8s as much as posible to setup everything).

Oh OK that'd be better. Does it have a default_tags awareness?

sorah:
Ensuring tagging is crucial for cost allocation (I need it for external project tho, but it's same in Cookpad also) so if the module creates a karpenter-related k8s manifest then we expect to inherit default_tags to make sure resources are properly tagged.

jportasa:
> Oh OK that'd be better. Does it have a default_tags awareness?

we add default tags at terraform project level, example: https://github.com/cookpad/global-web-platform/blob/main/terraform/production/shared/eks/terraform.tf#L28

maybe there is a more suitable way to do it, I don't know.

jportasa:
Well the tags are hardcoded in karepenter, in this case we use 2 "nodegroups"
- https://github.com/cookpad/global-web-platform/blob/main/k8s/infrastructure/base/karpenter/provisioner-default.yaml#L41
- https://github.com/cookpad/global-web-platform/blob/main/k8s/infrastructure/base/karpenter/provisioner-on-demand.yaml#L46

sorah:
> maybe there is a more suitable way to do it, I don't know.

See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/default_tags 

Some resources like ASG/Launch templates need to use aws_default_tags to inherit (this is what this PR does). So if this tf module creates a karpenter configuration using k8s provider, then those k8s objects should read tags based on aws_default_tags  data source to inherit.

sorah:
oh ymls are hardcoded, so not provided by the module at all?

jportasa:
> oh ymls are hardcoded, so not provided by the module at all?

karpenter doesn't inherit anything from terraform, well it can't actually, so we hardcode the tags in the yaml

sorah:
understood!

jportasa:
thank you @sorah ! we will be aware that the nodes are correctly tagged

sorah:
yeah I was thinking that the module will create a karpenter configuration using kubernetes provider, so module users don't have to write yamls

PR #341, opened 2023-06-06
1.25

errm:
- Upgrades to k8s 1.25
- Moves to a much more minimal setup
  - ASG node groups are no longer supported
  - Karpenter must be used to provision nodes
    - Karpenter is run via fargate
  - All addons / cluster configuration is removed from this module.
    - Recommended to manage via flux or other gitops
  
#  📝 TODO

- Upgrade procedure / instructions!

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -111,7 +111,7 @@ variable "kube_proxy_configuration_values" {
 
 variable "coredns_configuration_values" {
   type        = string
-  default     = null
+  default     = "{ \"computeType\": \"fargate\" }"
   description = "Configuration values passed to the coredns EKS addon."
 }
 


main.tf
```diff
@@ -49,60 +48,74 @@ resource "aws_cloudwatch_log_group" "control_plane" {
...
-  config  = local.config
-  replace = true
-  manifest = templatefile(
-    "${path.module}/storage_classes.yaml.tmpl",
-    {
-      provisioner = "ebs.csi.aws.com",
-      fstype      = "csi.storage.k8s.io/fstype: ${var.pv_fstype}",
-    }
-  )
+resource "kubernetes_config_map_v1_data" "aws_auth" {
```
jportasa, line 98: I see this block is very similar to `resource "kubernetes_config_map" "aws_auth"` why we have this twice?

main.tf
```diff
@@ -49,60 +48,74 @@ resource "aws_cloudwatch_log_group" "control_plane" {
...
+          rolearn  = aws_iam_role.karpenter_fargate.arn
+          username = "system:node:{{SessionName}}"
+          groups = [
+            "system:bootstrappers",
+            "system:nodes",
+            "system:node-proxier",
+          ]
+        },
+        {
+          rolearn  = aws_iam_role.karpenter_node.arn
```
jportasa, line 69: In web-platform-X  we use` instanceProfile: EKSNode` [here](https://github.com/cookpad/global-web-platform/blob/main/k8s/infrastructure/base/karpenter/provisioner-default.yaml#L35) we will have to change it in the provisioner?

karpenter_interruption_queue.tf
```diff
@@ -0,0 +1,74 @@
+resource "aws_sqs_queue" "karpenter_interruption" {
+  name                      = "Karpenter-${var.name}"
```
jportasa, line 2: maybe more clear karpepenter-spot-node-interruption-${var.name}?

variables.tf
```diff
@@ -0,0 +1,86 @@
+variable "name" {
```
jportasa, line 1: add a description

examples/cluster/main.tf
```diff
@@ -1,35 +1,32 @@
-terraform {
```
jportasa, line 1: create something like this https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/versions.tf

versions.tf
```diff
@@ -1,3 +1,15 @@
 terraform {
   required_version = ">= 1.0"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 4.47.0"
```
jportasa, line 7: 5.1.0 maybe?

versions.tf
```diff
@@ -1,3 +1,15 @@
...
+    }
+
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.10"
```
jportasa, line 12: 2.21.1 is the latest

karpenter_fargate.tf
```diff
@@ -0,0 +1,39 @@
+resource "aws_eks_fargate_profile" "karpenter" {
+  cluster_name           = aws_eks_cluster.control_plane.name
+  fargate_profile_name   = "Karpenter-${var.name}"
+  pod_execution_role_arn = aws_iam_role.karpenter_fargate.arn
+  subnet_ids             = values(var.vpc_config.private_subnet_ids)
+
+  selector {
+    namespace = "karpenter"
+    labels    = {}
```
jportasa, line 9: we want as well flux-system namespace to run on Fargate

versions.tf
errm: Same

main.tf
errm: That is the intention, but we can add the old profile name to the map during transition.

karpenter_interruption_queue.tf
```diff
@@ -0,0 +1,74 @@
+resource "aws_sqs_queue" "karpenter_interruption" {
```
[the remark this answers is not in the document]
errm: I don't think it makes it clearer, this queue will have a number of different events - not only spot interruptions.

PR #344, opened 2023-06-21
Fix 1.25-rc1

jportasa:
Issues:
- repeated twice` "resolve_conflicts_on_update = "OVERWRITE""`


errm (commented):
Could you add some information about the issue that this is solving?

I didn't see any issue with this in my test clusters?

PR #350, opened 2023-10-11
Add a poka-yoke to prevent unintentional destroy

errm:
It's quite easy to make a config change that requires the cluster to be destroyed and recreated.  This is usually not what you want since the state of the cluster will be lost.

This config will prevent terraform from trying to apply a change that requires the cluster to be destroyed ... if that is really what you wanted to do you must set `prevent_destroy = false` first


errm:
Currently there is no way to do this without hardcoding `prevent_destroy = true` inside the module...

This would also make the module untestable (at least with our current cleanup strategy).

PR #371, opened 2024-10-24
Enable to use Karpenter v1 controller policy

coord-e:
This PR attaches a controller policy for Karpenter v1 in the karpenter module. The policy content is taken from https://github.com/aws/karpenter-provider-aws/blob/v1.0.6/website/content/en/preview/getting-started/getting-started-with-karpenter/cloudformation.yaml. I've added module variables `v1` and `v1beta` to help migration as you did in https://github.com/cookpad/terraform-aws-eks/pull/368.

- policy JSON diff extracted from the cloudformation.yaml: https://gist.github.com/coord-e/04ad1c4bc54b60899e7b4f2b33bf07dd#file-policies-diff
- policy document Terraform diff extracted from this PR: https://gist.github.com/coord-e/04ad1c4bc54b60899e7b4f2b33bf07dd#file-documents-diff

I'll make a PR to backport this to v0.29 and v0.30 after this is merged.

### DIFF
--- a/modules/karpenter/controller_iam.tf
+++ b/modules/karpenter/controller_iam.tf
@@ -29,7 +29,7 @@ data "aws_iam_policy_document" "karpenter_controller_assume_role_policy" {
 }
 
 resource "aws_iam_role_policy" "karpenter_controller_v1_beta" {
-  count  = var.v1beta ? 1 : 0
+  count  = var.v1 ? 0 : 1
   name   = "KarpenterController-v1beta"
   role   = aws_iam_role.karpenter_controller.id
   policy = data.aws_iam_policy_document.karpenter_controller_v1_beta.json
--- a/modules/karpenter/controller_iam.tf
+++ b/modules/karpenter/controller_iam.tf
@@ -29,7 +29,7 @@ data "aws_iam_policy_document" "karpenter_controller_assume_role_policy" {
 }
 
 resource "aws_iam_role_policy" "karpenter_controller_v1_beta" {
-  count  = var.v1 ? 0 : 1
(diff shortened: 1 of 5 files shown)


errm (commented):
Could you check reason for test failure?

coord-e:
> LimitExceeded: Maximum policy size of 10240 bytes exceeded for role Karpenter-terraform-aws-eks-testing-3n8AYX

Including both v1 and v1beta policies inline in the role seems to exceed the IAM role's policy size limit. 
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html

This can be resolved by using managed policies. However, there should be no reason to attach both, as the v1 policy is strictly more restrictive than the v1beta policy (there should be no operations allowed by the v1 policy that aren't allowed by the v1beta policy, and Karpenter v1 should work with the v1beta policy). Therefore, it would be better to make them mutually exclusive.

Making v1 = false by default to avoid breaking changes, while allowing module users to set v1 = true if they wish to verify that their configuration works with v1 policies before v1beta is dropped.

coord-e:
> LimitExceeded: Maximum policy size of 10240 bytes exceeded for role Karpenter-terraform-aws-eks-testing-3n8AYX

Including both v1 and v1beta policies inline in the role seems to exceed the IAM role's policy size limit. 
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html

I resolved this with managed policies, now tests are passing 7257bc8e08b3007616b73c3b950124d2b006024c

errm (approved):
LGTM!

coord-e:
thank you!

PR #372, opened 2024-10-30
Use inline policy for v1beta to avoid resource recreation

coord-e:
follow-up of #371 

In #371, I separated Karpenter v1beta and v1 policies to their own managed policies. However I noticed that when we apply this change, the destruction of `aws_iam_role_policy` and creation of `aws_iam_role_attachment` can happen in any order, possibly result in a moment that no policy is attached to the role... it's sufficient to make only v1 a managed policy to address the original issue, so I revert the v1beta policy back to an inline policy in this PR to prevent the kind of disruption mentioned above.

### DIFF
--- a/modules/karpenter/controller_iam.tf
+++ b/modules/karpenter/controller_iam.tf
@@ -28,16 +28,16 @@ data "aws_iam_policy_document" "karpenter_controller_assume_role_policy" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


hfm (approved):

nekketsuuu (approved):

mozamimy (approved):

errm (approved):
LGTM

PR #386, opened 2025-05-01
Scan config with trivy

errm:


.trivyignore
nekketsuuu, line 1: Memo: After .trivyignore.yaml is GA, we can use it for specifying paths to be ignored https://trivy.dev/v0.62/docs/configuration/filtering/#trivyignoreyaml

nekketsuuu (approved):
Nice 👍 👍 👍 Thank you!

hfm (approved):

